PR #2, opened 2021-06-03
Fix air intake hatch errors

GlodBlock:


Dream-Master:
funny that i cant add more than one reviewer

src/Java/gtPlusPlus/xmod/gregtech/api/metatileentity/implementations/GT_MetaTileEntity_Hatch_AirIntake.java
```diff
@@ -210,7 +210,7 @@ public boolean isAirInHatch() {
 				return true;
 			}
 		}		
-		return false;
```
Glease, line 213: did you realize you just turned this function from sometimes false to always true?

Glease (changes requested):

src/Java/gtPlusPlus/xmod/gregtech/api/metatileentity/implementations/GT_MetaTileEntity_Hatch_AirIntake.java
GlodBlock: no, this function should only work when it is not empty, so it won't be always true.
if false, this hatch won't work anymore because "empty" is always false.

Glease: Look, there are only two return statements in the method body. Both of them have a literal `true` in it. How could it return false?

GlodBlock: oh no. i will add something else.

Glease (approved):

boubou19:
is this PR correcting fluid dupe with the air intake hatch?

Glease:
Yes

GlodBlock:
yes

PR #4, opened 2021-06-11
Disable log methane and rubber centrifuging

Prometheus0000:
Moving to GT5U completely


Dream-Master (changes requested):
i guess this one need to be touched too
https://github.com/GTNewHorizons/GTPP/blob/master/src/Java/gtPlusPlus/xmod/gregtech/loaders/misc/WoodCentrifuging.java#L66-L82

Prometheus0000:
None of it gets called anyway? That's where I started. Anyway, I just removed the whole thing like you wanted.

Dream-Master (approved):

PR #9, opened 2021-07-01
Stop checking for update as we are on our own now

Glease:


repo-alt (approved):
Deleting stuff, feels good.

PR #12, opened 2021-07-24
Fix GUI and WAILA fluid handling for GT++ single-block boilers and steam condenser.

### LINKED ISSUE #5644: [GT++] Advanced boiler [lv] is not capped at 32 buckets of steam
2.0.8.3
#
delta
#

see https://github.com/alkcorp/GTplusplus/issues/645

D-Cysteine:
I'm new to modding, so apologies in advance if I did something wrong or missed something. I was also unable to test the GT++ solar boiler and GT++ solar generator, because I could not find them in NEI. Maybe they are disabled?

Related pull request:
 * https://github.com/GTNewHorizons/GT5-Unofficial/pull/605

Bug fixes
---
 * Fixes incorrect GUI and WAILA steam capacity display (was half the correct amount) for GT++ single-block boilers.
 * Fixes water not being displayed by WAILA for the GT++ steam condenser.
 * Fixes steam not being displayed in the GUI for the GT++ steam condenser (steam bar, the leftmost one, is permanently empty).
   * The error was due to a line that set the computed steam capacity to `0`, due to it calling `GT_MetaTileEntity_BasicTank.maxSteamStore()`. I believe that this method is intended to be used by steam-*consuming* machines, such as the steam forge hammer, and so it is not overridden and returns `0` for the steam condenser, which is a steam-*producing* machine.
   * https://github.com/GTNewHorizons/GTPP/blob/b9e536405eef4a83a7746f87f590f9c22e17fa3c/src/Java/gtPlusPlus/xmod/gregtech/api/gui/CONTAINER_SteamCondenser.java#L61
 * Fixes GTNewHorizons/GT-New-Horizons-Modpack#5644

Changes
---
 * Updated GT++ code to handle my API change to `GT_Container_Boiler` and `GT_GUIContainer_Boiler`.
   * It looks like it's just the GT++ solar boiler which was affected, and I was unable to test it because it doesn't show up in NEI for me. Perhaps it's disabled?
 * Updated GregTech lib jar to one that includes my API change to `GT_Container_Boiler` and `GT_GUIContainer_Boiler`.
   * I'm new to modding and not sure that I did this right. Please double-check. Though, at least it builds locally for me just fine.

API changes
---
 * `int aSteamCapacity` parameter removed from several constructors.
   * This parameter was only used to set the displayed maximum steam capacity, and did not actually modify the machine's steam capacity. Being able to set this parameter is confusing, because using any value here besides the one that's actually used in internal calculations (i.e. `getSteamCapacity()`) is just going to cause the GUI to show the wrong value.
   * For the solar generator, this just looked like dead code. The steam capacity value was not getting read anywhere.
 * `int maxStorage` public field removed from `CONTAINER_AdvancedBoiler`.
   * This field was used to store the water capacity, and seemed to be unreferenced outside of one method, so I just made it a local variable.
 * `int mSteamCapacity` public field removed from `CONTAINER_SteamCondenser`.
   * This field was bugged and always set to 0. It also seemed to be unreferenced outside of one method, so I just made it a local variable.
 * I searched on GitHub and it looks like nothing depends on these particular APIs. If there actually are any other repositories which use this API, then they must be updated as well or this change will break them. Alternatively, I could add the parameter back in.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -147,7 +147,7 @@ dependencies {
     //compile files('libs/gregtech-5.09.31-dev.jar')
     //compile files('libs/gregtech-5.09.33-dev.jar')
     //compile files('libs/gregtech-5.09.34.17-dev.jar')
-    compile files('libs/gregtech-5.09.36.00-dev.jar')
+    compile files('libs/gregtech-5.09.37.00-dev.jar')
 
     provided "net.industrial-craft:industrialcraft-2:${config.ic2.version}:dev"  
     //compile "ic2:IC2Classic:1.2.1.8:dev" - Does not mirror internal class structure or visibility of IC2, won't ever compile.
--- a/build.gradle
+++ b/build.gradle
@@ -146,7 +146,8 @@ dependencies {
     //compile files('libs/gregtech-5.08.33-dev.jar')
     //compile files('libs/gregtech-5.09.31-dev.jar')
     //compile files('libs/gregtech-5.09.33-dev.jar')
-    compile files('libs/regtech-5.09.34.17-dev.jar')
+    //compile files('libs/gregtech-5.09.34.17-dev.jar')
+    compile files('libs/gregtech-5.09.36.00-dev.jar')
 
     provided "net.industrial-craft:industrialcraft-2:${config.ic2.version}:dev"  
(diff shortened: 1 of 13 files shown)


Glease (approved):
Seems good.

PR #13, opened 2021-07-25
Fix fluid pipe scaling

### LINKED ISSUE #8161: Inconsistent GT Fluid Pipe "Fluid Capacity"
2.1.1.0
#
Private server
#
Capacity progression in most fluid pipes works like this:
Tiny - Small - Normal - Large - Huge, 
which (seems to be due to their costs in ingots) have relative capacities of
1 - 2 - 6 - 12 - 24. 
Example (Bronze fluid pipes): 400 - 800 - 2400 - 4800 - 9600
#
Some of the materials do not adhere to this "rule":
- Trinium Naquadah Carbonite
- Staballoy
- Tantalloy (both)
- Void Metal
- Europium
- Potin
- Maraging Steel (both)
- Inconel (both)
- Hastelloy-X
- Tungsten
- Dark Steel
- Clay
- Lead

These follow linear progression (1 - 2 - 3 - 4 - 5), which makes little sense in relation to their material costs and encourages the use of many parallel small pipe pipelines. This is especially noticeable in early game with Potin pipes recommended by the Quest Book.
#
Correct the inconsistency.

D-Cysteine:
This was a pretty easy change, so I just went ahead and did it. See attached issue for discussion.
This change results in a significant capacity increase for regular-sized and larger GT++ fluid pipes.

 * Fixes GTNewHorizons/GT-New-Horizons-Modpack#8161

### DIFF
--- a/build.properties
+++ b/build.properties
@@ -4,6 +4,6 @@ ic2.version=2.2.817-experimental
 gt.version=5.09.31
 ae2.version=rv3-beta-22
 
-gtpp.version=1.7.06
+gtpp.version=1.7.07
 
 commit.hash=aa9f04218e5298414f900b9fe61131f7ed26b8f1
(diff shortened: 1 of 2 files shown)


Dream-Master:
glee8e — Yesterday at 9:55 PM
#13 should be fine

PR #17, opened 2021-08-01
Advanced Boilers

TimeConqueror:
Rewritten Advanced Boiler:
- Moved to abstract methods from GT Boiler.
- Now it stays around 100C to be able to generate steam, so it doesn't consume extra fuel for higher temperature.
- Just to clearify (was before): Consuming fuels with burnTime > 500 provides bonus to the temperature (so the fuel consuming rate will be slower)
- Removed explosion, because it's a smart boiler and nothing at temperature 100C can cause explosion. I just won't generate steam.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -110,7 +110,6 @@ build
 config
 crash-reports
 gradle
-libs
 local
 logs
 modpack
@@ -143,3 +142,6 @@ wip
 *.bat
 GT++.ipr
 GT++.iws
+
+
+out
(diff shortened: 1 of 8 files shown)


repo-alt:
Seems to be ok, did you actually test if it works?

TimeConqueror:
> Seems to be ok, did you actually test if it works?

Yep, I didn't find any bugs

repo-alt (approved):

PR #20, opened 2021-08-07
Update GT jars

D-Cysteine:
Starting up Minecraft with the latest versions of GT and GT++ results in a crash: https://pastebin.com/aRtSLkE6
Updating the jars fixes this crash.


TimeConqueror (approved):

D-Cysteine:
Oops never mind, I think I was using an outdated GT++ jar earlier when I was getting the error. Just rebuilt the latest one (without this PR) and it worked fine.

PR #24, opened 2021-08-14
Update Forestry to new API

D-Cysteine:
Unfortunately had to add Forestry to `libs/` since the version in the repo seems to be obfuscated, and won't compile with GT++


glowredman:
Is it neccessary to have the modified Forestry version? Otherwise you can just pull it from its maven: http://maven.ic2.player.to/net/sengir/forestry/forestry_1.7.10/4.2.16.64/ .

D-Cysteine:
GTNH Forestry 4.4.0.0 introduces a breaking change. GT++ in its current state is not affected by the change in API, but I figured it was better to update the version in case someone adds code later that is affected; if they don't build against the new version then everything will build fine but break in-game.

Dream-Master (approved):

PR #27, opened 2021-09-26
Disable autocrafter

kuba6000:
Disable autocrafter to prevent duplication bug and overall bad design.
@Dream-Master


Prometheus0000:
I don't think people will be happy with that. And if you're going to disable it, you should mention it in the quest. Also, if you're disabling it, why bother changing TT text?

Pxx500:
> I don't think people will be happy with that. And if you're going to disable it, you should mention it in the quest. Also, if you're disabling it, why bother changing TT text?

There is a very serious duplication bug with this machine I'm not sure whether I should tell how to replicate it, only crafter is disabled tho so I don't think people will be that angry

boubou19:
send me or any team member on discord in private the way of duplicating stuff so we can try to fix it instead of disabling it.

Pxx500:
@boubou19 were you able to replicate the duplication bug?

GTNH-Colen (approved):

boubou19 (approved):

PR #32, opened 2021-10-21
fix crush with the same name

GlodBlock:
see this https://github.com/GTNewHorizons/GT5-Unofficial/pull/697


mitchej123 (changes requested):
This will make nei not compatible with any version of gtpp that isnt shipped with gtnh. My goal is for broad 1.7.10 compat. If you want to change the names in our fork, please have gtpp register it's handlers with imc.

GlodBlock:
@mitchej123  done now

mitchej123 (approved):
Pending confirmation from Dream that it works.  Otherwise looks fine, it won't break the NEI global compat now :)

Dream-Master:
will check ingame

PR #42, opened 2021-11-18
updated build.gradle with new URLs, removed wireless charging local chat spam

bombcar:
replaced spaces with tabs
added some files to the gitignore


bombcar:
Currently as a draft - I can't get gt++ to build without these changes but @draknyte1 says they shouldn't be necessary.

I blame mojang

bombcar:
Can someone tell me if the build changes are correct/needed? Perhaps it would build without them now.

build.gradle
```diff
@@ -1,23 +1,24 @@
...
 		maven {
-			name = "gt"
-			url = "https://gregtech.overminddl1.com/"
-		}	
-        maven {
-            name = "sonatype"
-            url = "https://oss.sonatype.org/content/repositories/snapshots/"
-        }
+			name = "jitpack"
+			url = "https://jitpack.io"
```
glowredman: What do you need Jitpack for? Forge is pulled from maven.minecraftforge.net (which you defined below)

bombcar: I don't actually know, I was just copying things in until it built - but looks like that wasn't needed.

bombcar: hmm how do I indicate that 1142e97 below resolves this?

GlodBlock (approved):

PR #43, opened 2021-11-20
Tree Growth Simulator balancing + structurelib

Sphyix:
Give an alternative source for wood by re-enabling the GT++ multiblock.
Other wood farms are a lot more laggy in comparison to the multiblock. Growing trees normally, especially with a world accelerator, can have a noticeable impact on the server.

I tried balancing it with the expected wood output of an enderio farm with spruce saplings and world accelerators.
javaw_judcy4o5SM

World accelerator was placed on top of the farming station. Test with HV was done without a capacitor and test with ZPM was done with octadic capacitor.

EIO farm output, 1 minute:
3 amp hv (hv world accelerator) 2500 wood
3 amp zpm (zpm world accelerator) 6400 wood

GT++ Tree Growth Simulator, 1 minute:
1 amp hv 950 wood
1 amp uv 6500 wood

### DIFF
--- a/src/Java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntityTreeFarm.java
+++ b/src/Java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntityTreeFarm.java
@@ -17,6 +17,7 @@ import gregtech.api.interfaces.tileentity.IGregTechTileEntity;
 import gregtech.api.items.GT_MetaGenerated_Tool;
 import gregtech.api.metatileentity.implementations.*;
 import gregtech.api.objects.GT_RenderedTexture;
+import gregtech.api.util.GT_ModHandler;
 import gregtech.api.util.GT_Multiblock_Tooltip_Builder;
 import gregtech.api.util.GT_Recipe;
 import gregtech.api.util.GT_Utility;
@@ -53,8 +54,6 @@ public class GregtechMetaTileEntityTreeFarm extends GregtechMeta_MultiBlockBase 
 		CASING_TEXTURE_ID = TAE.getIndexFromPage(1, 15);
 		mCasingName = ItemUtils.getLocalizedNameOfBlock(ModBlocks.blockCasings2Misc, 15);
 	}
-
-
 
 /*
 	 * Static thread for Fake World Handling
@@ -279,31 +278,8 @@ if (executor == null || mTreeData == null) {
 		if (aTileEntity == null) {
 			return false;
 		} else {
-			IMetaTileEntity aMetaTileEntity = aTileEntity.getMetaTileEntity();
-			if (aMetaTileEntity instanceof GT_MetaTileEntity_Hatch_InputBus){
-				((GT_MetaTileEntity_Hatch)aMetaTileEntity).updateTexture(aBaseCasingIndex);
-				return this.mInputBusses.add((GT_MetaTileEntity_Hatch_InputBus)aMetaTileEntity);
-			} else if (aMetaTileEntity instanceof GT_MetaTileEntity_Hatch_Maintenance){
-				((GT_MetaTileEntity_Hatch)aMetaTileEntity).updateTexture(aBaseCasingIndex);
-				return this.mMaintenanceHatches.add((GT_MetaTileEntity_Hatch_Maintenance)aMetaTileEntity);
-			} else if (aMetaTileEntity instanceof GT_MetaTileEntity_Hatch_Energy){
-				((GT_MetaTileEntity_Hatch)aMetaTileEntity).updateTexture(aBaseCasingIndex);
-				return this.mEnergyHatches.add((GT_MetaTileEntity_Hatch_Energy)aMetaTileEntity);
-			} else if (aMetaTileEntity instanceof GT_MetaTileEntity_Hatch_OutputBus) {
-				((GT_MetaTileEntity_Hatch) aMetaTileEntity).updateTexture(aBaseCasingIndex);
-				return this.mOutputBusses.add((GT_MetaTileEntity_Hatch_OutputBus) aMetaTileEntity);
-			} else if (aMetaTileEntity instanceof GT_MetaTileEntity_Hatch_Muffler) {
-				((GT_MetaTileEntity_Hatch) aMetaTileEntity).updateTexture(aBaseCasingIndex);
-				return this.mMufflerHatches.add((GT_MetaTileEntity_Hatch_Muffler) aMetaTileEntity);
-			} else if (aMetaTileEntity instanceof GT_MetaTileEntity_Hatch_Input) {
-				((GT_MetaTileEntity_Hatch) aMetaTileEntity).updateTexture(aBaseCasingIndex);
-				return this.mInputHatches.add((GT_MetaTileEntity_Hatch_Input) aMetaTileEntity);
-			} else if (aMetaTileEntity instanceof GT_MetaTileEntity_Hatch_Output) {
-				((GT_MetaTileEntity_Hatch) aMetaTileEntity).updateTexture(aBaseCasingIndex);
-				return this.mOutputHatches.add((GT_MetaTileEntity_Hatch_Output) aMetaTileEntity);
-			}
+			return addToMachineList(aTileEntity, aBaseCasingIndex);
 		}
-		return false;
 	}
 
 	public int getMaxEfficiency(final ItemStack aStack) {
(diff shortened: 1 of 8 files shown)


src/Java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntityTreeFarm.java
```diff
@@ -240,35 +241,69 @@ public int getEuDiscountForParallelism() {
...
-								Logger.WARNING("Bad centrifuge casing");
-								return false;
-							}
-							++tAmount;
-
-						}
-					}
-				}
+			IMetaTileEntity aMetaTileEntity = aTileEntity.getMetaTileEntity();
+			if (aMetaTileEntity instanceof GT_MetaTileEntity_Hatch_InputBus){
```
draknyte1: Why do this?
There's a function in the base multi for it. addToMachine List iirc.

draknyte1 (changes requested):

src/Java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntityTreeFarm.java
```diff
@@ -25,16 +30,19 @@
...
 import net.minecraft.item.ItemStack;
-import net.minecraftforge.common.util.ForgeDirection;
+
+import static com.gtnewhorizon.structurelib.structure.StructureUtility.*;
+import static gregtech.api.util.GT_StructureUtility.ofHatchAdder;
 
 public class GregtechMetaTileEntityTreeFarm extends GregtechMeta_MultiBlockBase {
 
 	public static int CASING_TEXTURE_ID;
 	public static String mCasingName = "Advanced Cryogenic Casing";
```
draknyte1, line 46: Should be mCasingName = ItemUtils.getLocalizedNameOfBlock(ModBlocks.blockCasings2Misc, 15)

src/Java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntityTreeFarm.java
```diff
@@ -99,21 +100,28 @@ public String getMachineType() {
...
+		GT_Multiblock_Tooltip_Builder tt = new GT_Multiblock_Tooltip_Builder();
+		tt.addMachineType(getMachineType())
+				.addInfo("Converts EU to Oak Logs")
+				.addInfo("Eu Usage: 100% | Parallel: 1")
+				.addInfo("Requires a Saw or Chainsaw in GUI slot")
+				.addPollutionAmount(getPollutionPerTick(null) * 20)
+				.addSeparator()
+				.beginStructureBlock(3, 3, 3, true)
+				.addController("Front center")
+				.addCasingInfo("Sterile Farm Casing", 10)
```
draknyte1, line 118: Isn't localized, hence why mCasingName is set at the start of the tooltip. As the original variable will always be set to the unlocalized name. (Intended implementation)

src/Java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntityTreeFarm.java
```diff
@@ -300,27 +335,27 @@ public void onPostTick(IGregTechTileEntity aBaseMetaTileEntity, long aTick) {
...
 
 
 
 				if (!didElectricDamage && invItem.getItem() instanceof GT_MetaGenerated_Tool) {
 					long aDmg = GT_MetaGenerated_Tool.getToolDamage(invItem);
 					long aDmgMax = GT_MetaGenerated_Tool.getToolMaxDamage(invItem);
 					if (aDmg < aDmgMax && GT_MetaGenerated_Tool.getPrimaryMaterial(invItem) != Materials._NULL) {
-						Logger.WARNING("dmg: "+aDmg+" | max: "+aDmgMax);
+						//Logger.WARNING("dmg: "+aDmg+" | max: "+aDmgMax);
 						GT_MetaGenerated_Tool.setToolDamage(invItem, aDmg+getDamageToComponent(invItem));
```
draknyte1: Probably better to use GT_ModHandler.damageOrDechargeItem()?

draknyte1 (changes requested):

src/Java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntityTreeFarm.java
Sphyix: mCasingName is set in the constructor. That's doing the same thing.
I guess I can remove it from the constructor and set it directly.

src/Java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntityTreeFarm.java
Sphyix: Not sure what you mean, i'm not used of programming java and/or at gt++ code.
I only deleted a logger here. It looked like something used to debug the machine while code was being written.

src/Java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntityTreeFarm.java
Sphyix: I copied the implementation of the other multiblocks.
If there's a better way of doing this, please explain and I'll try to fix it

src/Java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntityTreeFarm.java
Sphyix: I tried using that but it was not working properly.
So I copied the implementation that was done on the rest of the multiblocks, unlocalized.

src/Java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntityTreeFarm.java
draknyte1: Use addToMachineList(hatch, textureID); instead of this ugly if else block.

src/Java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntityTreeFarm.java
draknyte1: I’m suggesting you change the way items are damaged to the method I’ve provided. It’s better.

src/Java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntityTreeFarm.java
Sphyix: Done, thanks.

src/Java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntityTreeFarm.java
Sphyix: I tested that method, minimum damage that can be done is 100.
Tool can't be easily replaced from buses or by pushing into the controller, so I preferred disabling the damage of it completely for now.
After testing on zeta we'll see if it's worth keeping like this or changing it back to damaging the tool.

draknyte1 (changes requested):
Tool damage should not be removed.

Sphyix:
> Tool damage should not be removed.

That's your opinion.
Adding consumables in the input bus and/or adding a fluid hatch with water could make sense, but (my opinion) I don't like the tool that needs to be replaced by hand, or the tool like that in general.

Florexiz:
Bonsai crop doesn't require any consumables, ender io farm can use unbreakable tools
There are no reason to use tools here imo

Glease:
Sounds like turbines and large turbine to me. How is that an issue?

repo-alt:
Can it use electric buzzsaw?

Sphyix:
> Sounds like turbines and large turbine to me. How is that an issue?

Mh yeah it's very similar.
Issue here is other machines would be easier to use and/or maintenance free compared to this multiblock.
Being a multiblock there is already maintenance on it, adding a consumable like this could make it too annoying to use compared to alternatives.

With the turbines there is no alternative, so not much can be done about them needing to be replaced by hand.

The only reason I'm trying to get this to work is to reduce the amount of tile entities with bonsai crops, and the tree spam with enderio farms.
If those 2 are easier to use and produce the same amount of wood, players won't use the multiblock.

So there are 2 things that can be done. Re-enabling the damage to the tool and buffing the production to make it more attractive, or leave it as it is now.

Sphyix:
> Can it use electric buzzsaw?

From my tests, no.
I don't really see a reason to add that anyway since there's no way to automatically recharge tools inside the machine, and the only good thing about buzzsaw and chainsaw is the power usage instead of durability.

It makes them not worth it considering the added cost if they can't be recharged. 

-edit-
Chainsaw seems to work fine

draknyte1:
> That's your opinion.

And I made the multi years ago. Unlike the other options you list, these scale and crops or other tree farms don’t. Having a consumable in the controller balances that. Later game you have tools with insane durability so it balances itself.

Sphyix:
> > That's your opinion.
> 
> And I made the multi years ago. Unlike the other options you list, these scale and crops or other tree farms don’t. Having a consumable in the controller balances that. Later game you have tools with insane durability so it balances itself.

It makes sense for lategame, I personally wouldn't have to change the tool at all after making a neutronium one.
But at lower tiers, it is annoying to keep changing tools, so the alternative is used.

Anyway I can put it back in, I'd like to have the multi on the next dev release for tests and balancing it better if needed.

Florexiz:
Okay, let just return tools for zeta test to see if its too annoying

draknyte1:
You can make it at HV at latest and not requiring a tool makes it instantly better than the alternatives. By this point you can still make vanadium steel tools and powered tools last even longer. There’s really no excuse. 

Bonsais in the forestry farm need the IC2 crop module iirc and that’s Uranium. EIO farmer wouldn’t even be viable if not the the unbreakable mattock quest reward.

Florexiz:
But you cant recharge tools there automatically
I prefer some other balancing like fluids in volcanus for example

Also you have ic2 crop harvesters that are now obtainable at lv or mv iirc
Unbreakable tools isn't only from quests, you can made them by yourself

src/Java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntityTreeFarm.java
```diff
@@ -240,34 +240,45 @@ public int getEuDiscountForParallelism() {
...
-		int tAmount = 0;
-		if (!aBaseMetaTileEntity.getAirOffset(xDir, 0, zDir)) {
+	@Override
+	public IStructureDefinition<GregtechMetaTileEntityTreeFarm> getStructureDefinition() {
+		if (STRUCTURE_DEFINITION == null) {
+			STRUCTURE_DEFINITION = StructureDefinition.<GregtechMetaTileEntityTreeFarm>builder()
+					.addShape(mName, transpose(new String[][]{
+							{"XXX", "CCC", "CCC"},
+							{"X~X", "C-C", "CCC"},
+							{"XXX", "CCC", "CCC"},
```
GlodBlock: just curious, why you don't allow to put hatches on the front face, while the tooltip says 'Any Casing'?

Sphyix: Good question, reply is a bit more dumb, I copy pasted from the centrifuge and left it as it was.
I'll change it to accept from the front aswell.

draknyte1: The centrifuge should also allow on the front, otherwise all mine will break. They face sideways, but are still technically on the front.

Sphyix: Pretty sure it was done to avoid this situations. That's gonna break half of my base aswell.
javaw_dPSmP3yVsK

draknyte1: If people want ugly broken textures centrifuges, let them. Doesn’t effect functionality and it still happens if they wallshare.

Sphyix: I agree tbh

draknyte1: @GlodBlock thoughts?

draknyte1 (approved):
Looks better now. If it works, awesome. 😎

src/Java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntityTreeFarm.java
GlodBlock: my eyes will bleed 😵

PR #45, opened 2021-11-24
fix catalyst durability bar

bombcar:
in attempting to work out why the catalyst doesn't get destroyed at 1 (or 0) I found the fix for the durability bar only going halfway.

### DIFF
--- a/src/Java/gtPlusPlus/plugin/agrichem/item/algae/ItemAgrichemBase.java
+++ b/src/Java/gtPlusPlus/plugin/agrichem/item/algae/ItemAgrichemBase.java
@@ -240,8 +240,8 @@ public class ItemAgrichemBase extends Item {
 				createCatalystNBT(aStack);
 	        }
 			double currentDamage = getCatalystDamage(aStack);
-			double durabilitypercent = currentDamage / 100;		
-	        return  durabilitypercent;
+			double durabilitypercent = currentDamage / getCatalystMaxDamage(aStack);
+	        return durabilitypercent;
 		}
 		else {
 			return 1D;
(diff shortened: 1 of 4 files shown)


Glease:
Maybe mark it as draft if you are not yet done, or people would come to review, only to get caught by another commit, and another commit, and another commit...

bombcar:
> Maybe mark it as draft if you are not yet done, or people would come to review, only to get caught by another commit, and another commit, and another commit...

I'll do that next time, this however I think is complete now ...

src/Java/gtPlusPlus/plugin/agrichem/item/algae/ItemAgrichemBase.java
```diff
@@ -240,8 +240,8 @@ public double getDurabilityForDisplay(ItemStack aStack) {
 				createCatalystNBT(aStack);
 	        }
 			double currentDamage = getCatalystDamage(aStack);
-			double durabilitypercent = currentDamage / 100;		
-	        return  durabilitypercent;
+			double durabilitypercent = currentDamage / getCatalystMaxDamage(aStack);
```
D-Cysteine, line 243: Hmm, just to be consistent with the other file, maybe make this line:
```
return currentDamage / getCatalystMaxDamage(aStack);
```

D-Cysteine (approved):
Just curious: was removing the gray text for >80% durability just a stylistic thing (i.e. green looks better)?

If you've tested this, then it looks good to me.

bombcar:
> Just curious: was removing the gray text for >80% durability just a stylistic thing (i.e. green looks better)?
> 
> If you've tested this, then it looks good to me.

Green -> yellow -> red makes more sense to me than gray when full, but I could go back to gray for 50 and green 49 and below

D-Cysteine:
I don't have a strong opinion either way; was just curious if it was fixing a bug or just a stylistic change

PR #46, opened 2021-11-25
fix chemplant not consuming catalyst

### LINKED ISSUE #6308: Chem. Plant catalyst usage
2.0.9.0QF4

#
Server Eta

#
 I wanted to make purple rocket fuel in one chemical plant with 8 parallels, but the following problems arose.
 1. The orange metal catalyst breaks during any other recipe in which it should not be used.
 2. When the strength of the catalyst reaches 1/50, it remains in this state and does not collapse
completely. But at the same time, empty catalysts appear in the ME. In the morning I threw 8 orange catalysts, and now I have 500 empty catalysts.
 3. And everything would not be so bad, but when the strength of the catalyst is in the 1/50 state, the chemical plant stops working in 8 parallels, but only in 1, which is VERY bad.

#
image
image
image

#
 Make sure that the catalyst does not lose strength when a recipe occurs in which it is not used.
Destroy the catalyst completely in a 1/50 strength state so that the new catalyst replaces it. This will immediately fix the problem with 1 active parallel instead of 8.

bombcar:
fix GTNewHorizons/GT-New-Horizons-Modpack#6308

The catalyst was being set null but it wasn't being synced because updateSlots doesn't touch the Catalyst Hatch.

I couldn't figure out how to make it do that, so I iterate over the array (which will only ever have one item) and update the slots.

Considering this is the first line of Java I've ever written, it's probably horribly wrong. *but it works*

Also removed an errant duplication of code.

### DIFF
--- a/src/Java/gtPlusPlus/xmod/gregtech/api/metatileentity/implementations/nbthandlers/GT_MetaTileEntity_Hatch_NbtConsumable.java
+++ b/src/Java/gtPlusPlus/xmod/gregtech/api/metatileentity/implementations/nbthandlers/GT_MetaTileEntity_Hatch_NbtConsumable.java
@@ -128,7 +128,6 @@ public abstract class GT_MetaTileEntity_Hatch_NbtConsumable extends GT_MetaTileE
 				fillStacksIntoFirstSlots();
 			}
 		}
-		tryFillUsageSlots();
 	}
 
 	// Only moves items in the first four slots
@@ -142,7 +141,7 @@ public abstract class GT_MetaTileEntity_Hatch_NbtConsumable extends GT_MetaTileE
 		}                
 	}
 	
-	private final void tryFillUsageSlots() {
+	public final void tryFillUsageSlots() {
 		int aSlotSpace = (mInputslotCount - getContentUsageSlots().size());
 		if (aSlotSpace > 0) {
 			Logger.INFO("We have empty usage slots. "+aSlotSpace);
--- a/src/Java/gtPlusPlus/xmod/gregtech/api/metatileentity/implementations/nbthandlers/GT_MetaTileEntity_Hatch_NbtConsumable.java
+++ b/src/Java/gtPlusPlus/xmod/gregtech/api/metatileentity/implementations/nbthandlers/GT_MetaTileEntity_Hatch_NbtConsumable.java
@@ -40,13 +40,6 @@ public abstract class GT_MetaTileEntity_Hatch_NbtConsumable extends GT_MetaTileE
 		mAllowDuplicateUsageTypes = aAllowDuplicateTypes;
 	}
 
-	public GT_MetaTileEntity_Hatch_NbtConsumable(String aName, int aTier, int aInputSlots, String[] aDescription, boolean aAllowDuplicateTypes, ITexture[][][] aTextures) {
-		super(aName, aTier, aInputSlots*2, aDescription[0], aTextures);
-		mInputslotCount = getInputSlotCount();
-		mTotalSlotCount = getInputSlotCount()*2;
-		mAllowDuplicateUsageTypes = aAllowDuplicateTypes;
-	}
-
 	@Override
 	public abstract ITexture[] getTexturesActive(ITexture aBaseTexture);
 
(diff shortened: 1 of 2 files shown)


src/Java/gtPlusPlus/xmod/gregtech/api/metatileentity/implementations/nbthandlers/GT_MetaTileEntity_Hatch_NbtConsumable.java
```diff
@@ -120,7 +120,8 @@ public final void updateSlots() {
 			if (i <= getSlotID_LastInput()) {
 				fillStacksIntoFirstSlots();
 			}
-		}            
+		}
+		tryFillUsageSlots();
```
bombcar: Not happy about this, but it's `private` and it calls updateSlots() I think ... which is this. But it works, it correctly loads the next catalyst, and it does NOT load two or something like that.

src/Java/gtPlusPlus/xmod/gregtech/api/metatileentity/implementations/nbthandlers/GT_MetaTileEntity_Hatch_NbtConsumable.java
```diff
@@ -40,13 +40,6 @@ public GT_MetaTileEntity_Hatch_NbtConsumable(String aName, int aTier, int aInput
 		mAllowDuplicateUsageTypes = aAllowDuplicateTypes;
 	}
 
-	public GT_MetaTileEntity_Hatch_NbtConsumable(String aName, int aTier, int aInputSlots, String[] aDescription, boolean aAllowDuplicateTypes, ITexture[][][] aTextures) {
```
bombcar, line 43: this is literally duplicate of the function above

draknyte1: It’s a constructor. They wouldn’t be identical.

draknyte1 (changes requested):
Removing constructors is bad.

src/Java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/chemplant/GregtechMTE_ChemicalPlant.java
```diff
@@ -766,6 +766,7 @@ public boolean checkRecipeGeneric(
 		this.mOutputItems = tOutputItems;
 		this.mOutputFluids = tOutputFluids;
 		updateSlots();
+		for (GT_MetaTileEntity_Hatch_Catalysts h : mCatalystBuses) h.updateSlots();
```
draknyte1: Keep the formatting consistent. 
new line with {}.

bombcar:
> Removing constructors is bad.

What's a constructor?

src/Java/gtPlusPlus/xmod/gregtech/api/metatileentity/implementations/nbthandlers/GT_MetaTileEntity_Hatch_NbtConsumable.java
bombcar: constructor good 
intelij bad

src/Java/gtPlusPlus/xmod/gregtech/api/metatileentity/implementations/nbthandlers/GT_MetaTileEntity_Hatch_NbtConsumable.java
Glease: Just turn tryFillUsageSlots into public. It won't hurt.

bombcar: ok I can do that

bombcar: Reverting because that didn't work .... the bug came back

bombcar: wait maybe I have the dumb

Dream-Master:
@draknyte1 is this ready for a merge ?

Dream-Master:
bombcar — Today at 5:12 PM
it works for me and I've tested it preeeety extensively
I say merge it, alks' comment was resolved but I couldn't mark it as such

Dream-Master (approved):

PR #53, opened 2021-11-28
GT++ fluids in NEI bookmarks fix

### LINKED ISSUE #18: GT++ bookmarked fluids changed after restart
Specifically, the GT++ fluids grisium and talonite, changed to titansteel and (Really long fission lithium fuel thing). After restarting they became two new GT++ fluids. I also confirmed that the entries in the bookmarks.ini file stayed the same, so I'd guess GT++ is registering it's fluids under a different metadata number every restart.

GTNH 2.0.8.5dev
NotEnoughItems-1.7.10-2.0.0-beta-6-GTNH-universal

I don't see any commits that would have fixed this in NotEnoughItems-1.7.10-2.0.1-GTNH-universal though.

slprime:
fix https://github.com/GTNewHorizons/NotEnoughItems/issues/18

p.s. about changes in build.gradle i dont sure becouse java is not my main language

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -165,9 +165,9 @@ dependencies {
 	//compile "ic2:IC2Classic:1.2.1.8:dev" - Does not mirror internal class structure or visibility of IC2, won't ever compile.
 
 	compile "mcp.mobius.waila:Waila:1.5.11-RC2-NONEI_1.7.10:dev"
-	compile "codechicken:CodeChickenLib:1.7.10-1.1.3.140:dev"
-	compile "codechicken:CodeChickenCore:1.7.10-1.0.7.47:dev"
-	compile "codechicken:NotEnoughItems:1.7.10-1.0.5.120:dev"
+	compile "com.github.GTNewHorizons:CodeChickenLib:master-SNAPSHOT:dev"
+	compile "com.github.GTNewHorizons:CodeChickenCore:master-SNAPSHOT:dev"
+	compile "com.github.GTNewHorizons:NotEnoughItems:2.1.16-GTNH:dev"
 	compile "api.player:PlayerAPI:1.7.10-1.4"
 	
 	compile "minetweaker:MineTweaker3:1.7.10-3.0.10:dev-full"
--- a/src/Java/gtPlusPlus/nei/FluidStackStringifyHandler.java
+++ b/src/Java/gtPlusPlus/nei/FluidStackStringifyHandler.java
@@ -1,0 +1,57 @@
+package gtPlusPlus.nei;
+
+import codechicken.nei.api.IStackStringifyHandler;
+import cpw.mods.fml.common.registry.GameData;
+import net.minecraft.nbt.NBTTagCompound;
+import codechicken.nei.NEIClientConfig;
+import net.minecraft.item.ItemStack;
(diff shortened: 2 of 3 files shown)


draknyte1 (changes requested):
Don't update build.gradle in your PR if you've made changes that aren't needed.
To my knowledge it was already updated in a previous PR.

Also why is this required? Should it not be in Gregtech itself?

slprime:
1. Without changes build.gradle mod don`t build
2. This fix gt++ fluids, for gregtech it not needed

bombcar:
> * Without changes build.gradle mod don`t build
> * This fix gt++ fluids, for gregtech it not needed

Try the latest build.gradle - yours may have been outdated. If it doesn’t build let’s find out why not just change things directly

draknyte1:
If your PR has nothing to do with the build file, just don’t push the changes you made locally. 

I don’t see how this is required for one mod and not the other when I copied fluid addition code from gregtech. Plus this is for the GT fluid display item, nothing specific to GT++?

Dream-Master:
Hope we get the git bot soon to auto bump version each merged pr

slprime:
1. Current build.gradle use `codechicken:NotEnoughItems:1.7.10-1.0.5.120:dev` but only `com.github.GTNewHorizons:NotEnoughItems:2.1.16-GTNH:dev` have `IStackStringifyHandler` interface. this is reason why i changed this file.

2. Bookmarks save item string name + damage + nbt for saving item in file. For gregtech fluids enough name + damage, but gt++ fluids have dynamic damage (fluid id). This is reason why after restart gtnh in bookmarks show another fluids

Before & After load minecraft. First is GT++ fluid, second gregtech fluid.
image
image

bombcar:
try `	compile "codechicken:NotEnoughItems:2.1.16-GTNH:dev"` maybe as a one-line change?

bombcar:
Ah, yea, I see that doesn't find it. Hmm - up to build experts to decide.

In general pulling from GTNH for GTNH isn't a major issue iirc.

Dream-Master:
@slprime maybe split this in two pr one for code change one for gradle change

draknyte1:
Fluid IDs are dynamic and should *NOT* be used as a reference for things anyway.
You can use FluidRegistry.getRegisteredFluidIDs() to update IDs as required, but using IDs in general is advised against regardless.
https://github.com/GTNewHorizons/NotEnoughItems/issues/18#issuecomment-606971479

https://maven.minecraftforge.net/net/minecraftforge/forge/1.8.8-11.15.0.1618-1.8.8/forge-1.8.8-11.15.0.1618-1.8.8-changelog.txt
Build 1.8.8-11.15.0.1617-1.8.8:
	LexManos: Deprecate int IDs in FluidRegistry. Modders should only ever use the String name. Also add a 'friendly' exception when attempting to get an ID for a unregistered fluid. Closes #1374

slprime:
@draknyte1 This PR fixes it. NEI does not know that GT_FluidDisplayItem is gluid, but GT++ does.

Or exists another solution detect in NEI (without new dependencies) what item is fluid and get it name?

slprime:
ok... i try find another way to fix this issue

PR #57, opened 2021-12-03
Unified build script

SinTh0r4s:
It will not run in dev. It probably requires a core mod rewrite for that. Or did it ever run in dev?


bombcar:
I was working on trying to get runClient to work and got somewhere iirc. It can be run from eclipse in debug mode.

draknyte1:
It works fine in dev. 

Won’t moving the source and asset directories break open PRs? If yes, then this shouldn’t be merged for a while.

gradle.properties
```diff
@@ -0,0 +1,64 @@
...
+mixinPlugin =
+# Specify the package that contains all of your Mixins. You may only place Mixins in this package or the build will fail!
+mixinsPackage =
+# Specify the core mod entry class if you use a core mod. This class must implement IFMLLoadingPlugin!
+# This parameter is for legacy compatability only
+# Example value: coreModClass = asm.FMLPlugin + modGroup = com.myname.mymodid -> com.myname.mymodid.asm.FMLPlugin
+coreModClass = preloader.asm.Preloader_FMLLoadingPlugin
+# If your project is only a consolidation of mixins or a core mod and does NOT contain a 'normal' mod ( = some class
+# that is annotated with @Mod) you want this to be true. When in doubt: leave it on false!
+containsMixinsAndOrCoreModOnly = false
```
draknyte1, line 56: Shouldn’t this be true?

Glease: For some reason it's containsMixinsAndOrCoreMod**Only**, not containsMixinsAndOrCoreMod

SinTh0r4s: This is intentional! Some mods consit out of a core mod only. This option exists to cover these edge cases! Default is `false`

draknyte1:
>Won’t moving the source and asset directories break open PRs? If yes, then this shouldn’t be merged for a while.

Still curious about this. @SinTh0r4s

bombcar:
> 

Git is surprisingly smart …. Sometimes. Especially if you merge in the right order. 

But I agree it should be merged after the big pr

SinTh0r4s:
> > Won’t moving the source and asset directories break open PRs? If yes, then this shouldn’t be merged for a while.
> 
> Still curious about this. @SinTh0r4s

Git is very good at tracking the movement of files. It will handle this without issues, unless files are new. This is unlikely to provide issues during this transition.

SinTh0r4s:
Fixed run in dev issues and resolved conflicts. It is in a mergable state right now

draknyte1:
> unless files are new. 

Well, I have lots of new files in my PR. So this should wait then.

PR #59, opened 2021-12-06
copy over gui-select-circuit in gregtech

Glease:
breakthrough circuits require at least HV machine to be show up in the list, so it doesn't break any gating.

### DIFF
--- a/src/Java/gtPlusPlus/core/item/circuit/ItemAdvancedChip.java
+++ b/src/Java/gtPlusPlus/core/item/circuit/ItemAdvancedChip.java
@@ -1,22 +1,38 @@
 package gtPlusPlus.core.item.circuit;
 
-import java.util.List;
+import java.util.*;
+import java.util.function.BiFunction;
+import java.util.function.Predicate;
 
+import cpw.mods.fml.common.FMLCommonHandler;
 import cpw.mods.fml.common.registry.GameRegistry;
+import gregtech.api.GregTech_API;
+import gregtech.api.enums.GT_Values;
+import gregtech.api.gui.GT_GUIDialogSelectItem;
+import gregtech.api.interfaces.INetworkUpdatableItem;
+import gregtech.api.net.GT_Packet_UpdateItem;
+import gregtech.api.objects.XSTR;
+import gregtech.api.util.GT_LanguageManager;
+import gregtech.api.util.GT_Utility;
 import gtPlusPlus.core.lib.CORE;
 import gtPlusPlus.core.util.minecraft.ItemUtils;
 import net.minecraft.client.renderer.texture.IIconRegister;
 import net.minecraft.creativetab.CreativeTabs;
 import net.minecraft.entity.Entity;
 import net.minecraft.entity.player.EntityPlayer;
+import net.minecraft.entity.player.EntityPlayerMP;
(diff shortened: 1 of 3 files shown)


src/Java/gtPlusPlus/core/item/circuit/ItemAdvancedChip.java
```diff
@@ -133,6 +156,82 @@ public IIcon getIcon(ItemStack stack, int pass) {
...
+		}
+		stack.setItemDamage(meta);
+
+		return true;
+	}
+
+	@Override
+	public boolean onItemUse(ItemStack stack, EntityPlayer player, World world, int x, int y, int z, int side, float xOffset, float yOffset, float zOffset) {
+		// nothing on server side or fake player
+		if (player instanceof FakePlayer || !world.isRemote) return false;
```
draknyte1, line 181: Use gtPlusPlus.core.util.minecraft.PlayerUtils.isRealPlayer(EntityLivingBase) > FakePlayer instancing. (TC doesn't extend FakePlayer)

src/Java/gtPlusPlus/plugin/agrichem/item/algae/ItemBioChip.java
```diff
@@ -133,6 +157,82 @@ public IIcon getIcon(ItemStack stack, int pass) {
...
+		}
+		stack.setItemDamage(meta);
+
+		return true;
+	}
+
+	@Override
+	public boolean onItemUse(ItemStack stack, EntityPlayer player, World world, int x, int y, int z, int side, float xOffset, float yOffset, float zOffset) {
+		// nothing on server side or fake player
+		if (player instanceof FakePlayer || !world.isRemote) return false;
```
draknyte1, line 182: Same as previous comment.

draknyte1 (approved):
Looks good otherwise.

src/Java/gtPlusPlus/core/item/circuit/ItemAdvancedChip.java
Glease: I thought you have patches to make it a FakePlayer?

draknyte1: Why would you think that?
If there isn't a patch here, then there isn't one: https://github.com/GTNewHorizons/GTplusplus/tree/master/src/Java/gtPlusPlus/preloader/asm/transformers.

Glease: Weird. I recall you showing off a patch for this on discord before covid started. IIRC to fix TT fake player converting sandstone into fluorite.

draknyte1: I just used PlayerUtils.isRealPlayer(EntityLivingBase). Didn't do anything beyond that.

Glease: It appears thaumcraft itself doesn't actually use the broken FakeThaumcraftPlayer at all. It's properly using the FakePlayer acquried from FakePlayerFactory (like everyone else). Since we do fork TT maybe I'll just go patch TT instead.

PR #62, opened 2021-12-07
Fix typo in Zyngen tooltip

Florexiz:


bombcar (approved):
Looks correct based on structure definition

PR #64, opened 2021-12-12
move recipe away from the item generator

### LINKED ISSUE #7694: Rhodium Duplication Glitch
2.1.0.0
#
Using packager to make tiny rhodium dust into rhodium dust. 9 tiny dust items should yield 1 dust item and 4 small rhodium dust items should yield 1 dust item.
#
2 small rhodium dust items can make 2 dust items and 1 tiny rhodium dust item can make 2 dust items.
image
image

#
Delete the two above recipes.
#
#### What is your GTNH Discord username?
xyq

GlodBlock:
to fix https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/7694
these recipe now will be put a independent recipe generator

### DIFF
--- a/src/Java/gtPlusPlus/core/material/Material.java
+++ b/src/Java/gtPlusPlus/core/material/Material.java
@@ -8,6 +8,7 @@ import java.util.*;
 import gregtech.api.enums.Materials;
 import gregtech.api.enums.OrePrefixes;
 import gregtech.api.enums.TextureSet;
+import gregtech.api.util.GT_Utility;
 import gtPlusPlus.api.objects.Logger;
 import gtPlusPlus.api.objects.data.AutoMap;
 import gtPlusPlus.core.item.base.BaseItemComponent.ComponentTypes;
@@ -29,6 +30,7 @@ import net.minecraft.item.Item;
 import net.minecraft.item.ItemStack;
 import net.minecraftforge.fluids.Fluid;
 import net.minecraftforge.fluids.FluidStack;
+import net.minecraftforge.oredict.OreDictionary;
 
 public class Material {
 
@@ -722,16 +724,21 @@ public class Material {
 
 	public final ItemStack getComponentByPrefix(OrePrefixes aPrefix, int stacksize) {
 		String aKey = aPrefix.name();
-		Map<String, ItemStack> g =  mComponentMap.get(this.unlocalizedName);
+		String aOreName = aKey + this.unlocalizedName;
+		for (ItemStack tTarget : OreDictionary.getOres(aOreName)) {
+			if ("bartworks".equals(ItemUtils.getModId(tTarget)) || "gregtech".equals(ItemUtils.getModId(tTarget)))
+				return GT_Utility.copyAmount(stacksize, tTarget);
+		}
+		Map<String, ItemStack> g = mComponentMap.get(this.unlocalizedName);
 		if (g == null) {
 			Map<String, ItemStack> aMap = new HashMap<String, ItemStack>();
 			mComponentMap.put(unlocalizedName, aMap);
 			g = aMap;
-		}		
+		}
 		ItemStack i = g.get(aKey);
 		if (i != null) {
 			return ItemUtils.getSimpleStack(i, stacksize);
-		}		
+		}
 		else {
 			// Try get a GT Material
 			Materials Erf = MaterialUtils.getMaterial(this.unlocalizedName);
@@ -754,7 +761,6 @@ public class Material {
 			//Logger.MATERIALS("Unabled to find \"" + aKey + this.unlocalizedName + "\"");
 			return ItemUtils.getErrorStack(stacksize, (aKey + this.unlocalizedName+" x"+stacksize));
 		}
-		
 	}
 
 	final public Block getBlock(){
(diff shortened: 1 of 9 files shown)


src/Java/gtPlusPlus/core/material/Material.java
```diff
@@ -722,16 +724,21 @@ public final boolean requiresBlastFurnace(){
 
 	public final ItemStack getComponentByPrefix(OrePrefixes aPrefix, int stacksize) {
 		String aKey = aPrefix.name();
-		Map<String, ItemStack> g =  mComponentMap.get(this.unlocalizedName);
+		String aOreName = aKey + this.unlocalizedName;
+		for (ItemStack tTarget : OreDictionary.getOres(aOreName)) {
```
draknyte1: Why do this? It negates the point of the cache.

src/Java/gtPlusPlus/xmod/gregtech/loaders/RecipeGen_MetalRecipe.java
```diff
@@ -0,0 +1,116 @@
...
+            if (GT_Values.RA.addCutterRecipe(
+                    material.getRod(1),
+                    material.getBolt(4),
+                    null,
+                    (int) Math.max(material.getMass() * 2L, 1L),
+                    material.vVoltageMultiplier)){
+                Logger.WARNING("Cut Blot Recipe: "+material.getLocalizedName()+" - Success");
+            }
+            else {
+                Logger.WARNING("Cut Blot Recipe: "+material.getLocalizedName()+" - Failed");
```
draknyte1: This also is spelt wrong.

src/Java/gtPlusPlus/core/material/Material.java
draknyte1: You don't even both to cache the result either...

src/Java/gtPlusPlus/core/material/Material.java
```diff
@@ -37,6 +37,7 @@
 	public static final Set<Material> mMaterialMap = new HashSet<Material>();
 
 	public static final Map<String, Map<String, ItemStack>> mComponentMap = new HashMap<String, Map<String, ItemStack>>();
+	public static final Map<String, ItemStack> mOreDictItemCache = new HashMap<String, ItemStack>();
```
draknyte1, line 40: No.. You've still yet to elaborate why you're doing this.
We don't need to add a cache for OD names when you can just cache the result from OD itself.

src/Java/gtPlusPlus/core/material/Material.java
```diff
@@ -725,10 +726,19 @@ public final boolean requiresBlastFurnace(){
...
-		for (ItemStack tTarget : OreDictionary.getOres(aOreName)) {
-			if ("bartworks".equals(ItemUtils.getModId(tTarget)) || "gregtech".equals(ItemUtils.getModId(tTarget)))
-				return GT_Utility.copyAmount(stacksize, tTarget);
+
+		if (mOreDictItemCache.containsKey(aOreName)) {
+			return GT_Utility.copyAmount(stacksize, mOreDictItemCache.get(aOreName));
+		}
+		else {
+			for (ItemStack tTarget : OreDictionary.getOres(aOreName)) {
+				if ("bartworks".equals(ItemUtils.getModId(tTarget)) || "gregtech".equals(ItemUtils.getModId(tTarget))) {
```
draknyte1, line 735: Why specifically target BW or GT? Unification should handle this if and when required.

draknyte1 (changes requested):

src/Java/gtPlusPlus/core/material/Material.java
GlodBlock: it won't and then it will generate dupe recipe like 2 small pile dust -> 1 full dust or 1 ingot -> 1 block

src/Java/gtPlusPlus/core/material/Material.java
GlodBlock: wut?

src/Java/gtPlusPlus/core/material/Material.java
draknyte1: So it's a problem in recipe generation itself..?
This issue you're trying to fix (Albeit in the wrong place) doesn't occur without BW.

src/Java/gtPlusPlus/core/material/Material.java
draknyte1:
```
		String aKey = aPrefix.name();
		Map<String, ItemStack> g =  mComponentMap.get(this.unlocalizedName);
		g.put(aKey, u);
		mComponentMap.put(unlocalizedName, g);
```
You're not putting the results in the material map for the given prefix.

src/Java/gtPlusPlus/core/material/Material.java
GlodBlock: gt should has a recipe check when adding recipe, but it fails to check when gtpp and bw both exist. considering the bw material is the main stream in nh, so i made the change in gtpp to force it use bw items.

draknyte1: Fix the issue at the source, don't try butcher up code here.
You seem to prefer bandaids over proper fixes or doing half assed jobs.

GlodBlock: source is here. and what i am doing is the proper fix.

draknyte1: It's not, or the issue would occur with just GT & GT++.

> gt should has a recipe check when adding recipe, but it fails to check when gtpp and bw both exist.

You said yourself it's GT.

> doing is the proper fix.

No, you really aren't. You've tried to skip the material component cache entirely and just use OD results which have to be looked up every time.

Glease:
This is obviously a bug with bartworks. It's trying to be smart and cached stuff where it shouldn't. Changing GT++ is really just hiding the bug under the rug.

PR #69, opened 2021-12-14
Pikyonium64B slight material adjustment

ScriptedPiky:
Slightly adjusts them. Are these materials correct? Hope it didn't drastically alter too much. Is this even okay to do for the most part?


draknyte1:
Why?

The only aspect of it that *would potentially* be adjusted is the Ytterbium but that's another story.

ScriptedPiky:
I just thought of updating the materials within slightly or changing them up. Though for Ytterbium I don't know what the state of that is but am open to adjusting that if needed

bombcar:
See https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/9105 for some recent discussions

draknyte1:
Unless there’s a necessary reason to change an alloy, then no.

draknyte1:
Imma just close this as it doesn’t really server any purpose.

PR #70, opened 2021-12-15
fix 9 in 1 desc

bombcar:
Update the Nine in One to show all nine machines.

### DIFF
--- a/src/Java/gtPlusPlus/core/handler/BookHandler.java
+++ b/src/Java/gtPlusPlus/core/handler/BookHandler.java
@@ -38,7 +38,7 @@ public class BookHandler {
 						"First you need to craft the following things for a Thermal Boiler to Function: The Main Boiler Block, 20 Thermal Containment Casings, two Input Hatches, two Output Hatches, a bunch of different Tools and a Maintenance Hatch.",
 						"To begin the building, lay out the first 3x3 layer of Machine Casings on the ground (with a Hatch in the Middle), then place the Boiler Block facing outward in the middle of one of the 3m wide Sides.", 
 						"Now grab 3 other Hatches and place them on the remaining three 3m wide Sides also facing outwards. And now the four corners of the Machine need also a Machine Casing. There should only be a Hole left in the middle of the Cube.",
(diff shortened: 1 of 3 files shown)


draknyte1 (changes requested):
Why change the Lang file?

You didn’t update the book. It’s in BookHandler.

bombcar:
> Why change the Lang file?
> 
> You didn’t update the book. It’s in BookHandler.

I'll update the book, the Lang file was to get it to stop complaining about invalid lines

draknyte1:
> > Why change the Lang file?
> > You didn’t update the book. It’s in BookHandler.
> 
> I'll update the book, the Lang file was to get it to stop complaining about invalid lines

// are comments though?

bombcar:
InteliJ doesn't like them:

image

but is fine with #:

image

bombcar:
The book handler is already correct ?

image

draknyte1:
> InteliJ doesn't like them:
> 
> image
> 
> but is fine with #:
> 
> image

We don’t all use IntelliJ (I don’t even use my IDE for Lang files). Don’t think we should change things that your IDE is complaining about when others may not. 

Guess I did the book ages ago then.

bombcar:
If I'm reading the language specification correctly, only # and ! are considered comments, the rest just parse incorrectly:

> A natural line that contains only white space characters is considered blank and is ignored. A comment line has an ASCII '#' or '!' as its first non-white space character; comment lines are also ignored and do not encode key-element information. In addition to line terminators, this format considers the characters space (' ', '\u0020'), tab ('\t', '\u0009'), and form feed ('\f', '\u000C') to be white space.

https://docs.oracle.com/javase/8/docs/api/java/util/Properties.html#load-java.io.Reader-

But - I can revert

draknyte1:
I guess you should just remake this now all the files have been moved. It’s cleaner for commit history & won’t have merge conflicts. 

As for comments in the Lang files, MC doesn’t care. Java comments start with // and all the other Lang files also use //. Might as well just keep it consistent.

PR #94, opened 2022-01-23
Tree Growth Simulator different wood outputs

Sphyix:
Changed Tree Growth Simulator to output wood based on sapling in the input bus.
I couldn't figure out how to keep the old code for it, with the generation of wood from a fake world, so based on what it was outputting I kept the balance and made the output a fixed amount that scales on the tier.

Formula is 2x²-2x+5 where x is the tier, this is the output/sec

### DIFF
--- a/src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/processing/GregtechMetaTileEntity_IndustrialDehydrator.java
+++ b/src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/processing/GregtechMetaTileEntity_IndustrialDehydrator.java
@@ -75,7 +75,7 @@ public class GregtechMetaTileEntity_IndustrialDehydrator extends GregtechMeta_Mu
 				.addInfo("Upgraded overclocks reduce recipe time to 25% and increase EU/t to 400%")
 				.addPollutionAmount(getPollutionPerSecond(null))
 				.addSeparator()
-				.beginStructureBlock(3, 4, 3, true)
+				.beginStructureBlock(3, 5, 3, true)
 				.addController("Bottom Center")
 				.addCasingInfo(mCasingName, 10)
 				.addInputBus("Any Casing", 1)
--- a/src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntityTreeFarm.java
+++ b/src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntityTreeFarm.java
@@ -24,6 +24,7 @@ import gtPlusPlus.xmod.gregtech.common.blocks.textures.TexturesGtBlock;
 import gtPlusPlus.xmod.gregtech.common.helpers.TreeFarmHelper;
 import gtPlusPlus.xmod.gregtech.common.helpers.treefarm.TreeGenerator;
 import net.minecraft.init.Blocks;
+import net.minecraft.item.Item;
 import net.minecraft.item.ItemStack;
 
 import static com.gtnewhorizon.structurelib.structure.StructureUtility.*;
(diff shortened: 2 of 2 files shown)


draknyte1:
There is a much easier way to do this without a huge ugly switch statement.

You can get the OreDict name of the sapling, then get the log from OD by changing the String.
`saplingBirch -> logBirch, etc.`
This can then be cached, significantly improving performance.
Using a GT_ItemStack as a map key (Of the sapling) and storing the log as the value.
`sLogCache.put(new GT_ItemStack(aSapling), aLogStack);`
`ItemStack aOutput = sLogCache.get(new GT_ItemStack(aSapling); if (aOutput != null) { aOutput = aOutput.copy()}`

Or something like that.

draknyte1 (changes requested):
See above.

Sphyix:
Changed to HashMap instead of the switch.
I'm populating it manually since I don't see a pattern sapling->log that I can use, every mod is doing a different thing

draknyte1 (changes requested):
You don't ever need to manually declare mappings, you can just get the oreDict tags from the sapling, then just replace `sapling` with `log`, if it yields a valid return, then cache it.

Sphyix:
It can't work if I just replace sapling with log.
It works with minecraft saplings, gt++, barnarda c, but not the rest that I put in.
Like for example, if I try to replace log with sapling with the TConstruct slime
`TConstruct:slime.sapling` becomes `TConstruct:slime.log`, and that doesn't exist since the right block to output is `TConstruct:slime.gel`
Same thing applies to all the natura trees, let's take another example, let's say I want bloodwood
So  `Natura:florasapling` becomes `Natura:floralog`, that doesn't exist again, since bloodwood log is `Natura:bloodwood`

So I don't understand what you mean by just replacing sapling with log...

draknyte1:
Don't use full item names.

```java
	public static ItemStack getLogFromSapling(ItemStack aSapling) {
		if (aSapling != null) {
			int[] aOreDictIDs = OreDictionary.getOreIDs(aSapling);
			if (aOreDictIDs != null && aOreDictIDs.length > 0) {
				ArrayList<String> aOreDictTags = new ArrayList<String>();
				for (int aID : aOreDictIDs) {
					aOreDictTags.add(OreDictionary.getOreName(aID));
				}
				for (String aTag : aOreDictTags) {
					if (aTag.startsWith("sapling")) {
						String aLogTag = aTag.replace("sapling", "log");
						ItemStack aLogStack = ItemUtils.getItemStackOfAmountFromOreDictNoBroken(aLogTag, 1);
						if (ItemUtils.checkForInvalidItems(aLogStack)) {
							for (ItemStack aTestStack : OreDictionary.getOres(aLogTag, false)) {
								if (GT_Utility.areStacksEqual(aTestStack, aLogStack, false)) {
									return aLogStack.copy();
								}
							}
						}
					}
				}
			}
		}
		return null;
	}
```

Then you can use

```java
	public static HashMap<GT_ItemStack, ItemStack> sSaplingCache = new HashMap<GT_ItemStack, ItemStack>();
	
	public static boolean cache(ItemStack aSapling, ItemStack aLog) {
		if (aSapling == null || aLog == null) {
			return false;
		}
		return sSaplingCache.put(new GT_ItemStack(aSapling), aLog) == null;
	}
	
	public static ItemStack getLogs(ItemStack aSapling, int aAmount) {
		if (aSapling == null || aAmount < 1) {
			return null;
		}
		else {
			if (aAmount > 64) {
				aAmount = 64;
			}
			ItemStack aLogs =  sSaplingCache.get(new GT_ItemStack(aSapling));
			if (ItemUtils.checkForInvalidItems(aLogs)) {
				return ItemUtils.getSimpleStack(aLogs, aAmount);
			}
		}
		return null;
	}
```

boubou19:
the issue with that is that you make the assumption that all the sapplings and logs are registered through their dedicated oredict entries. But i didn't find yet any sapling that isn't registered to something different in addition to `treeSapling`, same for logs as they seem to all be registered as `logWood`

draknyte1:
> the issue with that is that you make the assumption that all the sapplings and logs are registered through their dedicated oredict entries. But i didn't find yet any sapling that isn't registered to something different in addition to `treeSapling`, same for logs as they seem to all be registered as `logWood`

I didn't even think of that. 🤦🏻‍♀️

Sphyix:
At least I learned something new, thanks :D
Now for this issue, how are we solving it?

Dream-Master:
@draknyte1 can you review this or let us know what need to be changed else

Dream-Master:
@Sphyix any news ?

Sphyix:
Guess hardcoding is the only option then.
Multi is working, tested.
If @draknyte1 is ok, this can be merged

Dream-Master:
@draknyte1

draknyte1 (approved):

PR #98, opened 2022-01-26
Add output hatch to Cyclotron (again)

Savallator:
Since some patch, the Cyclotron no longer allowed an Output hatch, but needs it to run some recipes.


draknyte1 (approved):
Glod broke it. x]

Savallator:
Can anyone explain me why, when i build master, i get 1.7.24, but in the pack there is 1.7.26? I can not find any branch producing that, and on jenkins there is also only 1.7.24...

mitchej123:
Make sure you do a `git fetch --all`  Versions are based on tags, and tags are lazily pulled I recall.

PR #101, opened 2022-01-28
Update buildscript and dependencies to allow runClient

bombcar:
runClient needs to have the libraries for NEI to be present so they are compile instead of compileOnly.


Dream-Master:
it builds so i guess it works :)

Dream-Master (approved):

PR #102, opened 2022-01-29
St00f

draknyte1:


src/main/java/gtPlusPlus/core/handler/events/PlayerSleepEventHandler.java
```diff
@@ -0,0 +1,106 @@
...
+import net.minecraft.potion.PotionEffect;
+import net.minecraft.util.ChatComponentTranslation;
+import net.minecraftforge.event.entity.player.PlayerSleepInBedEvent;
+import net.minecraftforge.event.entity.player.PlayerWakeUpEvent;
+
+public class PlayerSleepEventHandler {
+
+	private static Field sEffectDuration = ReflectionUtils.getField(PotionEffect.class, "duration");
+	private static ArrayList<Potion> sPositiveEffects = new ArrayList<Potion>();
+	private static ArrayList<Potion> sNegativeEffects = new ArrayList<Potion>();
```
mitchej123, line 25: Any reason not to just use a `Map<int, Potion>` and look it up instead of iterate over it for every active effect?

Suppose it's not that much iteration just... feels dirty.

src/main/java/gtPlusPlus/core/item/general/ItemMagicFeather.java
```diff
@@ -0,0 +1,269 @@
...
+			boolean hasItem = hasItem(player, ModItems.itemMagicFeather);
+			if (hasItem != this.hasItem) {
+				if (hasItem) {
+					this.onAdd();
+				}
+				if (!hasItem) {
+					this.onRemove();
+				}
+				this.hasItem = hasItem;
+				Logger.INFO("Ticking feather "+hasItem);
```
mitchej123, line 234: A2

mitchej123 (approved):
Gave it a quick glance over; seems reasonableish.

src/main/java/gtPlusPlus/core/item/general/ItemMagicFeather.java
Glease: GT++'s logger doesn't log anything unless you explicitly turn it on.

draknyte1: Heh, mark it as changes requested or dream will beat me to removing this.

PR #122, opened 2022-02-11
Super Glue Line

Steelux8:
- First draft of the full Super Glue line, an IV processing chain meant to add a higher tier glue, starting from glue-related chemicals such as Methyl Acetate, with another tier of glue in the works

This is a work-in-progress chain for a higher tier of glue. This chain is intended for IV, and as such I mean to add this Super Glue to IV and later recipes, with one example being the tier 2 Thermal Padding.

This is a very Chemical Plant-intensive chain, compared to most others that exist, and usually steers away from the LCR spam used in everything from Polyethylene to Polybenzimidazole. I've heard several developers comment on the overreliance on the LCR, so I created this to contribute to a shift towards other multiblocks. I'm of the opinion that the Chemical Plant is the right choice for mid and late-game chemistry recipes, given its size, complexity and options for customization. Here, for example, the chain requires multiple GT++ catalysts, including a custom one I made, and the pipe casing tier will be very important to maintain said catalysts.

All numbers are not final (ignore the DT recipe time, it'll definitely be adjusted). The EU and machine casing tiers are mostly as intended, although they can also be changed if it's better that way. Here are screenshots of the chain:

<details>
  <summary>Screenshots</summary>
  

  
</details>


draknyte1 (commented):
Looks good. 👍
Just mark it ready when you’re done.

GTNH-Colen:
I'm not really sure what the point of this one is, I feel it is only going to frustrate users if they are forced to use this for a few minor things (like the thermal padding). I do like the usage of the chem plant of LCR though.

draknyte1:
> I'm not really sure what the point of this one is, I feel it is only going to frustrate users if they are forced to use this for a few minor things (like the thermal padding). I do like the usage of the chem plant of LCR though.

So best we give it lots of uses x]

GTNH-Colen:
Sure but you are moving away from glue by the time you are making thermal padding (T2 rocket) so I'd like to understand more of where this is intended to go before adding a large chain for 1-2 items.

vovaksenov:
Now that manual maintenance is done easily with toolboxes, the braintech duct tape recipe can be changed to use the superglue. This will encourage people to automate this process if they want automaintenance tier 1.

Steelux8:
> I'm not really sure what the point of this one is, I feel it is only going to frustrate users if they are forced to use this for a few minor things (like the thermal padding). I do like the usage of the chem plant of LCR though.

The point of this chain is to fill a slot of a sorely lacking series of chemicals. Plastics have a whole progression, rubbers also have one, but glues do not have such a thing, to the extent where mixing methyl acetate with polyvinyl acetate, byproducts of the "biological" powergen chains, yields the same glue you get from sticky resin in the Steam age.

As for uses for this IV Super Glue, I thought of the following:
-> Fluid input for the maintenance tape, either required or yielding more tape than regular glue
-> T2 thermal padding fluid input, with PBI used to make the 4 parts still (T1 thermal padding could be changed to an MV glue, the acetate mix I mentioned above, after making that recipe output a better glue than the one it currently does)
-> Graphene fluid input, instead of normal glue
-> Binder for the Super Fuels that I revamped the advanced large boilers for, they only consume these fuels but output lots of superheated steam
-> Alternative input for advanced SMDs (either PBI or super glue), or for crafting the later tier hatches that need PBI, or the advanced batteries that need PBI

(In reality, a lot of recipes that use PBI could very well use an advanced glue, instead)

-> Fluid input to craft the quantumsuit, instead of Titanium
-> etc.

In reality, a lot of recipes that use a fluid to "stick things together" use molten plastics, and not glue. I see many possibilities to use this chain.

Dream-Master (approved):

PR #126, opened 2022-02-13
Put Multiblock Sifter to HV

Dream-Master:
image


GTNH-Colen:
Why though?

draknyte1:
Doesn't make any sense, it's still using a MV/HV material and this multi is not meant to be IV tier like the rest.

Dream-Master:
we can discuss to move it to HV but i want this device stay in one tier and use all items from this tier

draknyte1:
It already was HV...

Dream-Master:
Now the circuits hv as well so the cables

draknyte1 (approved):

PR #128, opened 2022-02-13
balancing Potin Pipes now between steel and stainless steel

Dream-Master:
image

1000 Liter/sec (tiny Pipes)


Dream-Master:
if this values to good or to bad i need suggestions but potin is bronze tin and lead.

GTNH-Colen (approved):

PR #132, opened 2022-02-15
update buildscripts

bombcar:


.github/scripts/test-no-error-reports.sh
```diff
@@ -0,0 +1,33 @@
...
+
+RUNDIR="run"
+CRASH="crash-reports"
+SERVERLOG="server.log"
+
+if [[ -d $RUNDIR/$CRASH ]]; then
+  echo "Crash reports detected:"
+  cat $RUNDIR/$CRASH/crash*.txt
+  exit 1
+fi
```
leagris: You need `nullglob` option to avoid returning the pattern itself when no match exist.
Here a better implementation to get the crash report:
```sh
# enable nullglob to get 0 results when no match rather than the pattern
shopt -s nullglob
# store matches in array (don't forget to double-quote variables expansion
crash_reports=( "$RUNDIR/$CRASH/crash"*.txt )
if [ "${#crash_reports[@]}" -gt 0 ]; then
  latest_crash_report="${crash_reports[-1]}"
  {
    printf 'Latest crash report detected %s:\n' "${latest_crash_report##*/}"
    cat "$latest_crash_report"
  } >&2
  exit 1
fi

File: .github/scripts/test-no-error-reports.sh [hunk 2]
```diff
@@ -0,0 +1,33 @@
...
+
+if grep --quiet "Fatal errors were detected" $SERVERLOG; then
```
leagris: double-quote variable expansion `"$SERVERLOG"`
Also `grep` raw text rather than glob pattern: with `grep -F`:
```sh
if grep --quiet --fixed-strings 'Fatal errors were detected' "$SERVERLOG"; then
```

.github/scripts/test-no-error-reports.sh
```diff
@@ -0,0 +1,33 @@
...
+  echo "Server force stopped:"
+  cat server.log
+  exit 1
+fi
+
+if ! grep --quiet -Po '.+Done \(.+\)\! For help, type "help" or "\?"' $SERVERLOG; then
+  echo "Server didn't finish startup:"
+  cat server.log
+  exit 1
+fi
```
leagris: Same issues with unquoted variable expansion, use of `echo` and no redirect to stderr.

leagris (changes requested):
The shell script part need fixing:
- VERY IMPORTANT: Unquoted variable expansions.
- IMPORTANT: Refactor of latest crash log capture in array with `nullglob` option.
- MEDIUM IMPORTANCE: Avoid using `echo`, replace with `printf`.
- IMPROVEMENT: Group and redirect error outputs to `stderr` with `>&2`.

.github/scripts/test-no-error-reports.sh
bombcar: these files come from ExampleMod and should be fixed there

.github/scripts/test-no-error-reports.sh
leagris: ok

leagris (approved):

Dream-Master (approved):

src/main/java/gtPlusPlus/core/lib/CORE.java
```diff
@@ -321,13 +321,13 @@ public static EntityPlayerMP getFakePlayer(World world) {
 	public static class Everglades{
 		public static final String MODID = "ToxicEverglades";
 		public static final String NAME = "GT++ Toxic Everglades";
-		public static final String VERSION = "0.1";
```
draknyte1, line 324: These are separate mods and shouldn’t be versioned the same as the main mod?

dependencies.gradle
```diff
@@ -1,48 +1,26 @@
 // Add your dependencies here
 
 dependencies {
-    compile("com.github.GTNewHorizons:GT5-Unofficial:5.09.40.32:dev")
-    compile("com.github.GTNewHorizons:StructureLib:1.0.15:dev") 
-    compile("com.github.GTNewHorizons:NotEnoughItems:2.2.6-GTNH:dev") 
-    compile("com.github.GTNewHorizons:CodeChickenCore:1.1.3:dev")
-    compile("com.github.GTNewHorizons:CodeChickenLib:1.1.5.3:dev")
-    compile("net.industrial-craft:industrialcraft-2:2.2.828-experimental:dev")
+    compile('com.github.GTNewHorizons:GT5-Unofficial:5.09.40.35:dev')
```
draknyte1, line 4: Why are deps bumped every time anyone does anything? I thought the point of explicit deps was so that didn’t happen unless the updated deps were actually required…

mitchej123: Things are still in flux and transitive deps are being added.  Updating or not updating, as long as the builds are reproducible, and the underlying API hasn't changed,  shouldn't matter in practice.

src/main/java/gtPlusPlus/core/lib/CORE.java
bombcar: I can drop these back - or since merge master already merged do you just want to update on your copy

dependencies.gradle
bombcar: Once the whole thing calms down we'll be able to leave them as is; but right now too many things are being changed (mainly the build scripts :( )

PR #138, opened 2022-02-25
control circuits should not be 0

Dream-Master:
Change recipes using Circuit 0 in their recipes #9818
* https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/9818

### DIFF
--- a/src/main/java/gtPlusPlus/core/recipe/RECIPES_GREGTECH.java
+++ b/src/main/java/gtPlusPlus/core/recipe/RECIPES_GREGTECH.java
@@ -1943,32 +1943,32 @@ public class RECIPES_GREGTECH {
 	private static void cyclotronRecipes() {
 
 		// Polonium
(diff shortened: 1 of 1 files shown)


draknyte1:
These are intentionally 0.. why change them?

Dream-Master:
Circuit 0 should be the default setting not used for any recipe

Dream-Master:
we have enough nummbers free. If not we can add circuit 25-48

draknyte1:
> Circuit 0 should be the default setting not used for any recipe

Maybe in your opinion, but as the one who added the recipes it was wholly intentional. Changing them because you don’t like the use of 0 is silly. 0 is a valid circuit number and there is no reason it can’t be used.

Elisis (approved):

Dream-Master:
its more because the circuit programmer not using 0 and logical there is no other mod use now circuit 0 we should change that.
I dont touche the code much only cosmetic things

PR #139, opened 2022-02-25
Fix Elemental dublicator tooltip

LewisSaber:


draknyte1 (approved):
I didn’t think I actually ever changed the structure check to use these, guess I did. 🥲

LewisSaber:
after few days of crafting i finally made it< and it didnt assemble
so i took all hatches and used structure blueprint on it and found out

PR #140, opened 2022-02-26
hydrofluoric acid dupe

GlodBlock:
chem plant: 2000 industry HF + 5000 HF = 6000 industry HF
chem reactor: 4000 industry HF + 8 Na dust = 8 NaF
chem reactor: 8NaF + 4H2SO4 = 4Na2SO4 + 8HF
chem reactor: 4Na2SO4 + 8H = 8Na + 4H2SO4
final: 4H = 3HF

### DIFF
--- a/src/main/java/gtPlusPlus/core/recipe/RECIPES_GREGTECH.java
+++ b/src/main/java/gtPlusPlus/core/recipe/RECIPES_GREGTECH.java
@@ -422,7 +422,7 @@ public class RECIPES_GREGTECH {
 
 					},
 					new FluidStack[] {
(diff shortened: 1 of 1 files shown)


Pnc6lx:
in this chain, 4000 industry HF = 8HF, so it should be 5500 in first line
edit: 2000 + 5000 / 2 = 4500

draknyte1:
1L if Industrial HF is worth 2L of HF. I don’t believe there should be ways to turn IHF directly into HF. Recipes should be balanced around this.

GlodBlock:
NJWP8WL2Z}GNY0~4CHT5(BU
let's see what players are doing to dupe hydrofluoric acid

PR #146, opened 2022-03-04
Changed EBF faster recipes to not have circuit 1s

Steelux8:
- Changed the ABS recipes taken from the EBF recipe map to ask for different circuits on the recipes that use a fluid input to reduce crafting time for specific dusts.

The ABS has some EBF recipes included in its recipe map, but the ones that use fluid inputs such as Oxygen to accelerate the crafting use the same circuit as the slower alternatives, circuit 1, and as such these recipes are never chosen by the ABS. With this change, the ABS uses the same circuit as the EBF for these, usually circuit 11, which allows the faster recipes to be used in the ABS.

### DIFF
--- a/src/main/java/gtPlusPlus/xmod/gregtech/loaders/RecipeGen_BlastSmelterGT_GTNH.java
+++ b/src/main/java/gtPlusPlus/xmod/gregtech/loaders/RecipeGen_BlastSmelterGT_GTNH.java
@@ -158,20 +158,40 @@ public class RecipeGen_BlastSmelterGT_GTNH {
 					}
 					//If this recipe is enabled and we have a valid molten fluidstack, let's try add this recipe.
 					if (enabled && isValid(inputs, outputs, inputsF, mMoltenStack)) {						
+						// Boolean to decide whether or not to create a new circuit later
+						boolean circuitFound = false;
+
 						//Build correct input stack
 						ArrayList<ItemStack> aTempList = new ArrayList<ItemStack>();
 						for (ItemStack aPossibleCircuit : inputs) {
 							if (!ItemUtils.isControlCircuit(aPossibleCircuit)) {
 								aTempList.add(aPossibleCircuit);
 							}
+							else {
+								aTempList.add(aPossibleCircuit);
+								circuitFound = true;
+							}
 						}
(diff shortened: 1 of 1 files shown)


src/main/java/gtPlusPlus/xmod/gregtech/loaders/RecipeGen_BlastSmelterGT_GTNH.java
```diff
@@ -158,20 +158,40 @@ public synchronized static boolean generateGTNHBlastSmelterRecipesFromEBFList()
 					}
 					//If this recipe is enabled and we have a valid molten fluidstack, let's try add this recipe.
 					if (enabled && isValid(inputs, outputs, inputsF, mMoltenStack)) {						
+						// Boolean to decide whether or not to create a new circuit later
+						boolean circuitFound = false;
+
 						//Build correct input stack
 						ArrayList<ItemStack> aTempList = new ArrayList<ItemStack>();
 						for (ItemStack aPossibleCircuit : inputs) {
 							if (!ItemUtils.isControlCircuit(aPossibleCircuit)) {
```
draknyte1:
```
	if (ItemUtils.isControlCircuit(aPossibleCircuit)) {
		circuitFound = true;
	}
	aTempList.add(aPossibleCircuit);
```

Probably a tad cleaner.

draknyte1 (changes requested):
Looks good otherwise.

src/main/java/gtPlusPlus/xmod/gregtech/loaders/RecipeGen_BlastSmelterGT_GTNH.java
Steelux8: Right, I forgot to clean this up after I checked it works. I also change the "aPossibleCircuit" variable name to just recipeItem because it confused me at first, I don't like it.

draknyte1 (approved):
@bombcar you don’t need to rerequest reviews. Could’ve just approved it yourself if the issue has been addressed. :p

bombcar:
> @bombcar you don’t need to rerequest reviews. Could’ve just approved it yourself if the issue has been addressed. :p

I like to re-request when it's something I don't fully understand 🤣

PR #156, opened 2022-03-11
Nerf PSS

Elisis:


draknyte1 (commented):
You should probably comment on what changed in the PSS logic.

Elisis:
Of course.

Loss increased to 5% the voltage of hatches from 2%
Passive loss increased from 0.1% to 1%
Redox cells have 4 tier-1 circuits and 2 tier-1 superconductor wires added to their recipes

draknyte1 (approved):

PR #161, opened 2022-03-28
Super Glue Fixes

Steelux8:
- Reworked recipe EU/t and casing tiers to divide the Chemical Plant recipes evenly between 3 tiers. The purpose of this change is to make it so this line can be automated with 3 plants, one of each kind;
- Changed Chemical Plant's minimum base outside casings to allow some more hatches before the multi fails to form. It's a large, smart multiblock with many casings that are never replaced with hatches, so I believe the amount of allowed hatches at once should be increased;
- A few fixes for recipes that were missing, or implemented incorrectly.

### DIFF
--- a/src/main/java/gtPlusPlus/xmod/gregtech/loaders/recipe/RecipeLoader_GlueLine.java
+++ b/src/main/java/gtPlusPlus/xmod/gregtech/loaders/recipe/RecipeLoader_GlueLine.java
@@ -12,6 +12,7 @@ import gregtech.api.util.GT_ModHandler;
 import gregtech.api.util.GT_OreDictUnificator;
 import gregtech.api.util.GT_Utility;
 import gtPlusPlus.api.objects.Logger;
+import gtPlusPlus.core.item.ModItems;
 import gtPlusPlus.core.item.chemistry.AgriculturalChem;
 import gtPlusPlus.core.item.chemistry.GenericChem;
 import gtPlusPlus.core.item.chemistry.NuclearChem;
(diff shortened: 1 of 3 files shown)


draknyte1 (approved):

src/main/java/gtPlusPlus/xmod/gregtech/loaders/recipe/RecipeLoader_GlueLine.java
```diff
@@ -224,28 +236,45 @@ private static void dehydratorRecipes() {
...
                 null,
                 4 * 20,
-                MaterialUtils.getVoltageForTier(4));
+                MaterialUtils.getVoltageForTier(5));
+    }
+
+    private static void electrolyzerRecipes() {
+        GT_Values.RA.addElectrolyzerRecipe(
+                CI.getNumberedCircuit(1),
+                MISC_MATERIALS.CALCIUM_CHLORIDE.getDust(1),
```
GlodBlock, line 258: why 1 CaCl2 dust can be electrolyzed into 1 Ca dust and 2000mb Cl? you should keep all materials process consist with others. also there's already CaCl2 material ingame, why register duplicate material?

Steelux8: I thought there was, but it must have been in a different mod that wasn't in my dev environment since I didn't see any when testing. If it already exists, I'll remove this recipe and new material and refer to that one instead. Thank you for letting me know.

PR #162, opened 2022-03-29
Remove Duplicate Calcium Chloride and Use BW's Instead

Steelux8:
Creating a Calcium Chloride material in GT++ is unnecessary, Bartworks already has it defined. If BW isn't loaded, the recipe that outputs this material will instead output Calcium.


Steelux8:
This PR has conflicts, I'll make another one.

PR #169, opened 2022-04-13
use the proper recipe input checker instead of always reinvent the wheel

Glease:
Fix https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/10091

PR #171, opened 2022-04-14
implement oredict cycling in NEI

Glease:
Most of GT++ multi should have the oredict capability. We can fix it later if some of them still doesn't support oredict.


miozune:
Is this an attempt to address issue mentioned in #169? I think we need to touch inputs, not outputs

draknyte1 (approved):

draknyte1 (commented):
Actually I didn’t notice you applied this to outputs 😂

miozune:
my bad, this is fine

PR #183, opened 2022-04-29
Add XL Large Boiler - Industrial Superheater

Steelux8:
- Groundwork for the Industrial Superheater Implementation, a LuV Steam Boiler multiblock that turns Super Fuels into Superheated Steam at a much faster rate than any of the Large Boilers;
- Added a new alloy, RPTh Steelux Alloy, to be used in the casings of this multi;
- Added recipes for the casings and controller (recipes are not final, still need changes);
-  Added custom textures for the casings.

Missing the logic and some fixes to structure lib, to properly define how many of each hatch are needed and to fix the textures of hatches when the multi is formed.

### DIFF
--- a/src/main/java/gtPlusPlus/core/handler/COMPAT_HANDLER.java
+++ b/src/main/java/gtPlusPlus/core/handler/COMPAT_HANDLER.java
@@ -166,6 +166,7 @@ public class COMPAT_HANDLER {
 			GregtechIndustrialRockBreaker.run();
 			GregtechIndustrialChisel.run();
 			GregtechIndustrialFluidHeater.run();
+			GregtechIndustrialSuperheater.run();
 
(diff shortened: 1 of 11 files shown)


LewisSaber:
Its cool but what was the point of adding even more alloys that are used only once

Steelux8:
I added an alloy because I wanted to, and because GT++ makes it easy to do so. Aside from that, it's used only once right now, since I just implemented it today. It's not the only use I want it to have.

Elisis:
Couldn't you just, y'know, use an existing alloy?

Steelux8:
I could, but I don't think there are any material ID limitations from GT++. I also want to use the alloy in other places, and it's the only one I've created, it's not a problem.

PR #186, opened 2022-05-04
Making the Algae Farm Viable

Steelux8:
- Buffed almost all related recipes by a lot, greatly boosting outputs and reducing inputs and craft time.

The Algae Farm is an early, tiered GT++ multiblock that generates algae at a very slow rate, going faster as it tiers up, generating them at no cost for as long as it runs. The tier is decided by a "floor" of GT machine casings of a certain tier, in a similar way to the chemplant. Depending on the tier, more or less types of algae can be obtained, and the whole set of them can be processed into a lot of different chemicals and items, like Ethylene/Polyethylene, Sticky Resin, Methanol, Charcoal, Ammonia and many others. 

However, the recipes are unbelievably expensible, and never worth considering. Here are some examples I mathed out quickly earlier, compared to basic ways of obtaining the same items:

1000 Algae Biomass -> 1000L Methane (160,000 EU/bucket) (Ignores the water)
1000 Algae Biomass + 25000L Dist. Water -> 1000L Methane (256,000 EU/bucket, 1000 seconds)
17 Wood -> 1020L Methane (68,000 EU/bucket, 170 seconds) (easiest recipe)

512 Green Algae Biomass + 256 Compost + 16000L Urea Mix -> 64 Fertilizer
512 Green Algae Biomass + 2560 Green Algae Biomass + 16000L Urea Mix -> 64 Fertilizer
3072 Green Algae Biomass + 16000L Urea Mix -> 64 Fertilizer (204,800 + 2,880,000 = 3,084,800 EU/stack, 3840 seconds)
(Above calc doesn't include the cost of Ammonia, etc. for Urea Mix)
64 Calcite Dust + 64 Dark Ashes -> 64 Fertilizer (192,000 EU/stack, 320 seconds) (easy recipe)

512 Cellulose Pulp + 9216L Liquid Resin -> 64 Sticky Resin (14,000,000+ EU/stack)
128 Rubber Wood -> 64 Sticky Resin (512,000 EU/stack) (easiest recipe)

As seen above, the Algae Farm processing takes much longer and eats up millions of EU to produce any amount of useful outputs. As such, I buffed almost all of them by a lot, so that they more closely match the rest of the recipes for the same outputs. The benefit of using this setup is that the resources keep spawning forever, like a void miner would do, but the main disadvantage is that both the Algae Farm and the Chemical Plant are needed for the full setup, two large multis which need dozens of tiered casings to work, each.

These are quick changes, although I tried to keep everything balanced. Golden Brown Algae are only obtainable at a low chance from HV onwards, and Red Algae from EV onwards. I'd like some feedback on this draft PR, so that this fully funcional multiblock and corresponding setup can find a place in the pack and add more diversity to an early-mid game player's options.

### DIFF
--- a/src/main/java/gtPlusPlus/plugin/agrichem/BioRecipes.java
+++ b/src/main/java/gtPlusPlus/plugin/agrichem/BioRecipes.java
@@ -214,17 +214,17 @@ public class BioRecipes {
 	private static void recipeGreenAlgae() {
 		// Compost
 		GT_ModHandler.addPulverisationRecipe(
-				ItemUtils.getSimpleStack(AgriculturalChem.mGreenAlgaeBiosmass, 10),
+				ItemUtils.getSimpleStack(AgriculturalChem.mGreenAlgaeBiosmass, 4),
 				ItemUtils.getSimpleStack(AgriculturalChem.mCompost, 1));
 
 		// Turn into Cellulose
 		CORE.RA.addSixSlotAssemblingRecipe(new ItemStack[] {
 				getGreenAlgaeRecipeChip(),
-				ItemUtils.getSimpleStack(AgriculturalChem.mGreenAlgaeBiosmass, 30)
+				ItemUtils.getSimpleStack(AgriculturalChem.mGreenAlgaeBiosmass, 10)
 		},
 				GT_Values.NF,
 				ItemUtils.getSimpleStack(AgriculturalChem.mCelluloseFiber, 5),
-				30 * 30,
+				5 * 30,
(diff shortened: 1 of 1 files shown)


draknyte1 (approved):
Looks good so far.

Steelux8:
If Alk approves it, I'll set it as ready to merge. I wanted to change the recipes to something closer to GTNH numbers, so I can now look at tuning the new numbers based on tiering, so that this setup can be an additional option to consider, without making other options irrelevant.

PR #189, opened 2022-05-16
Implementing the XL Heat Exchanger

Steelux8:
- Fixed the processing math so that this multi works as fast as 16 LHEs;
- Changed inner pipe casings to Tungstensteel to add to the cost of building it;
- Added recipes for controller and the special casings. They are LuV recipes that use Zeron-100, which needs an LuV Energy Hatch on the ABS;
- Clarified the structure requirements and processing speed on the tooltip.

GT++ parallel multi for the LHE, works as fast as 16 LHEs. Instead of using mostly Titanium, it uses a lot of Tungstensteel and an LuV GT++ alloy.


GlodBlock:
there is already a XHL in gg, no need to add another advanced heat exchanger

ScriptedPiky:
I'm not sure when the XLHE was added in GT++, but I assume it was technically before the one in GG but was rather obscure

Steelux8:
> there is already a XHL in gg, no need to add another advanced heat exchanger

Do you mean the Extreme Heat Exchanger? I did not think of it as a parallel multi to the LHE, because it outputs Supercritical Steam at max speed, which doesn't work on the turbines used for the LHE. I didn't see documentation on the power output/input flow of the SC Turbine based on the materials, either. To use SH Steam, the max speed of the EHE is 5x of the LHE, compared to the 16x of this XL LHE.

Additionally, now that you speak of it, the EHE is 10x faster than the LHE despite only needing some stacks of Tungstensteel instead of Titanium. That, to me, is absolutely ridiculous and very overpowered. However, looking at the recipe for the SC Turbine, I suppose the total cost of the setup is moved to those turbines? This is not simply doing LHE parallels as the XL LHE does.

Steelux8:
After talking with Glod about the other "XHL in gg", which is the Extreme Heat Exchanger, we agreed to a 32x multiplier instead of 16x, since the EHE is IV and runs 5x faster when outputting Superheated Steam, or 10x faster when outputting the new Supercritical Steam, which requires new and more expensive turbines to process.

GlodBlock (approved):

PR #190, opened 2022-05-20
Add LCR recipes for Platinum to the Chemical Plant

Steelux8:
- Added recipes so that one Stainless Steel, MV Machine Casing chemplant can do the plat part of the platline, excluding the EBF, Sifter and Electrolyzer recipes.

**This is not replacing the LCR recipes, just providing an alternative option.**

This is a "trial PR" for my attempt at making Chemical Plants more popular, as well as reducing LCR spam. Important chemical lines should be doable in chemplants, and I'd argue they could skip some steps as long as the inputs and outputs are the same. 

That is what I tried to do here. The only input change is water, which I removed because it's 100% free and renewable, to make Nitric Acid. The Platinum Dust is made directly from Platinum Concentrate by adding the Calcium Dust right away, instead of having it as two different recipes. On this recipe, the output is not Nitrogen Dioxide + Diluted Sulfuric Acid, but instead a direct Aqua Regia output. Aqua Regia needs Water, Oxygen, Nitrogen Dioxide and Diluted Sulfuric Acid to create, so I added Oxygen as an input, ignored the Water and replaced the other two with the Aqua Regia output.

Calcium Chloride still needs an Electrolyzer, Platinum Salt Dust still needs a Sifter and an EBF.

If this gets approved, I'll work on copying over more recipes to the Chemical Plant.


ScriptedPiky:
I'm curious but is the [line#82](https://github.com/GTNewHorizons/GTplusplus/pull/190/commits/f590d88021f43fc5827d8a2bfb16dd53ce7aa74a#diff-7f99b71ec161371a0b23d4e5872e1af6f5ba62c8643073a4d2b3185a977ad06fR82) `4*20,` the time required? In this case, this would be 4 seconds. Or is the time required the `400` on [line#83](https://github.com/GTNewHorizons/GTplusplus/pull/190/commits/f590d88021f43fc5827d8a2bfb16dd53ce7aa74a#diff-7f99b71ec161371a0b23d4e5872e1af6f5ba62c8643073a4d2b3185a977ad06fR83)?

Otherwise, I think this is pretty neat. Not sure on how this will have a direct impact for balance and or mid to late-game balance, but overall pretty neat

Steelux8:
4*20 is the amount of ticks per recipe, 400 is the EU/t.

jonanlsh:
I suggest that we could use this to expand the chemplant and make it a way to condense long processing chains into a single machine.

1) Have the chemplant take in all the inputs involved in a single processing chain and produce all the outputs that the entire processing chain would produce in the end. 

2) To facilitate 1, let the chemplant now accept input/output blocks on all 4 sides at both the top and bottom of the plant. This will allow those super-long later game processing chains to have viable chemplant alternatives available at higher tiers, vs when they are initially introduced. Eg: platline fully installed in IV, chemplant platline in LuV/ZPM

chochem:
I really like the idea of (optionally) replacing multiple LCR steps by a big chemplant. In this case parts of the platinum line. This should also be extended to the rest of the platline though a replacement of the osmium DT (if replaceable at all) should probably still be quite costly/slow.

The main change I would recommend is to tier all the recipes higher, for example in ZPM. A common theme in the pack/gregtech is that new tiers often give you access to new and (in some way) simpler recipes. This would fit in very well.

Steelux8:
I've thought about it, and I'll rework this concept so that the chemplants craft each element of the platline in one step, and tier it at lategame. However, there are not enough output slots to do this, so I'll have to find a way to output everything. I'll work on these changes.

Steelux8:
I will redo this PR in a more endgame, even less steps format, in a different PR.

PR #192, opened 2022-05-24
Added Uses to Cactus/Sugar Charcoal/Coke

Steelux8:
- Changed the existing Industrial Coke Oven recipes to make them a lot faster than the Charcoal recipes, since they only make 1 at a time compared to the Charcoal recipes' 5;
- Added uses to these items, such as Charcoal Byproducts output, Wood Tar extraction, Coal Tar and Gas, and also CO and CO2.

As per https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/10233, Charcoal and Coal Coke can be made from Cactus and Sugar Cane, although with less fuel value than the original versions. However, these don't have most of the uses that those do, being mostly used as furnace fuel and little else. With this PR, I changed and created recipes so that these can be used similarly to Wood and Charcoal, although with less available uses and automations, which leave Wood and Charcoal as the best options after the first couple of tiers.

### DIFF
--- a/src/main/java/gtPlusPlus/xmod/railcraft/HANDLER_Railcraft.java
+++ b/src/main/java/gtPlusPlus/xmod/railcraft/HANDLER_Railcraft.java
@@ -79,15 +79,15 @@ public class HANDLER_Railcraft {
 			CORE.RA.addCokeOvenRecipe(aInputs2[i], CI.getNumberedCircuit(4), FluidUtils.getFluidStack("nitrogen", 50), FluidUtils.getFluidStack("charcoal_byproducts", 100), aOutputs[i], 20, 16);
 
 			// Generate renewable Coal Tar and Coal Gas from these Cokes
-			CORE.RA.addCokeOvenRecipe(aOutputs[i], CI.getNumberedCircuit(5), FluidUtils.getFluidStack("steam", 100), FluidUtils.getFluidStack("fluid.coaltar", 200), null, 60, 240);
-			CORE.RA.addCokeOvenRecipe(aOutputs[i], CI.getNumberedCircuit(6), FluidUtils.getFluidStack("steam", 100), FluidUtils.getFluidStack("fluid.coalgas", 300), null, 60, 240);
+			CORE.RA.addCokeOvenRecipe(aOutputs[i], CI.getNumberedCircuit(5), FluidUtils.getFluidStack("steam", 100), FluidUtils.getFluidStack("fluid.coaltar", 200), Materials.Ash.getDustSmall(1), 60, 240);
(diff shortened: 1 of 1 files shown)


Steelux8:
Still waiting for approval on this to close the issue mentioned above.

src/main/java/gtPlusPlus/xmod/railcraft/HANDLER_Railcraft.java
```diff
@@ -67,8 +72,23 @@ private static void generateCokeOvenRecipes() {
...
+			CORE.RA.addCokeOvenRecipe(aInputs1[i], CI.getNumberedCircuit(3), null, FluidUtils.getFluidStack("creosote", 100), aInputs2[i], 20, 16);
+			CORE.RA.addCokeOvenRecipe(aInputs1[i], CI.getNumberedCircuit(4), FluidUtils.getFluidStack("nitrogen", 100), FluidUtils.getFluidStack("charcoal_byproducts", 200), aInputs2[i], 10, 16);
+			CORE.RA.addCokeOvenRecipe(aInputs2[i], CI.getNumberedCircuit(3), null, FluidUtils.getFluidStack("creosote", 200), aOutputs[i], 40, 16);
+			CORE.RA.addCokeOvenRecipe(aInputs2[i], CI.getNumberedCircuit(4), FluidUtils.getFluidStack("nitrogen", 50), FluidUtils.getFluidStack("charcoal_byproducts", 100), aOutputs[i], 20, 16);
+
+			// Generate renewable Coal Tar and Coal Gas from these Cokes
+			CORE.RA.addCokeOvenRecipe(aOutputs[i], CI.getNumberedCircuit(5), FluidUtils.getFluidStack("steam", 100), FluidUtils.getFluidStack("fluid.coaltar", 200), Materials.Ash.getDustSmall(1), 60, 240);
+			CORE.RA.addCokeOvenRecipe(aOutputs[i], CI.getNumberedCircuit(6), FluidUtils.getFluidStack("steam", 100), FluidUtils.getFluidStack("fluid.coalgas", 300), Materials.Ash.getDustSmall(1), 60, 240);
```
GlodBlock: remove coal tar/gas recipe or change them into wood tar/gas.

GlodBlock (changes requested):

src/main/java/gtPlusPlus/xmod/railcraft/HANDLER_Railcraft.java
Steelux8: It's done. I liked the renewable Coal Tar and Gas, but Coal itself is already renewable, and making the Wood versions from these makes more sense.

Dream-Master (approved):

PR #193, opened 2022-05-26
Added Bismuth Indium Solder Alloy

Steelux8:
- Added Indalloy 140, IV recipe, meant for Assembly Line recipes.

Assembly Line recipes use all kinds of high tier items to craft the machine components for LuV and above. In the middle of all that, most recipes use the LV-tier Soldering Alloy, since nobody else implemented a higher tier one. That is what this alloy is for, to be used as a solder in those recipes.

[This is a real soldering alloy, as seen here.](https://www.matweb.com/search/DataSheet.aspx?MatGUID=61beeec304154e14b711984757f073ed&ckck=1). It's made in the ABS, with one IV Energy Hatch being required. It's mostly Bismuth, but also requires Indium.

image

### DIFF
--- a/src/main/java/gtPlusPlus/core/item/ModItems.java
+++ b/src/main/java/gtPlusPlus/core/item/ModItems.java
@@ -638,6 +638,7 @@ public final class ModItems {
 
 			MaterialGenerator.generate(ALLOY.BABBIT_ALLOY, false);
 			MaterialGenerator.generate(ALLOY.BLACK_TITANIUM, false);
+			MaterialGenerator.generate(ALLOY.INDALLOY_140, false);
 
 			// High Level Bioplastic
 			MaterialGenerator.generate(ELEMENT.STANDALONE.RHUGNOR, false, false);
--- a/src/main/java/gtPlusPlus/core/material/ALLOY.java
+++ b/src/main/java/gtPlusPlus/core/material/ALLOY.java
@@ -968,6 +968,25 @@ public final class ALLOY {
 					new MaterialStack(ELEMENT.getInstance().ARSENIC, 2)
 			});
 
+	public static final Material INDALLOY_140 = new Material(
+			"Indalloy 140", //Material Name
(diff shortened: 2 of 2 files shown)


glowredman (commented):
Why are you implementing this in GT++? Wouldn't GT be a better place?

Steelux8:
> Why are you implementing this in GT++? Wouldn't GT be a better place?

It's meant to be crafted in the ABS, and GT++ is much better to define alloys in anyways.

glowredman (approved):

GlodBlock:
ABS isn't a good place to make these alloy in large quantities. it only sopports one energy hatch so it will become ass to overclock it, people have to spamming it in mid game. give it a mixer recipe

Steelux8:
> ABS isn't a good place to make these alloy in large quantities. it only sopports one energy hatch so it will become ass to overclock it, people have to spamming it in mid game. give it a mixer recipe

It makes 100 ingots of the alloy every 40 seconds at IV speed. That seems more than fast enough to me, since it'll be overclocked 3 times by the time the player is UV. Even being a 2x/4x, that's 20 ingots per second.

GlodBlock:
20 ingots per second is far from enough, soldering alloy is used everywhere. abs recipe don't have parallel like ebf recipe, unless you add a mega abs.

GTNHAfx237v7:
I can only speak for myself but I keep soldering alloy stocked in a tank. You can set up passive production for this, hence I believe 100ingots every 40 seconds is fine. If you want to go stargate-crazy you can always setup multiple ABS or store that alloy in quantum tanks/bedrockium drums etc.
Apart from that a Mega ABS could be interesting but that is not the point of this ticket.

Steelux8:
> 20 ingots per second is far from enough, soldering alloy is used everywhere. abs recipe don't have parallel like ebf recipe, unless you add a mega abs.

1 ingot/tick is not enough for assline recipes? What needs faster soldering alloy than that?

PR #195, opened 2022-05-30
Comment out duplicate keys

MuXiu1997:
Comment out duplicate keys

reason: In some scripts used for localization, the key is expected to be unique

- context: https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/10419

### DIFF
--- a/src/main/resources/assets/miscutils/lang/en_US.lang
+++ b/src/main/resources/assets/miscutils/lang/en_US.lang
@@ -55,11 +55,11 @@ item.MiscUtils.material.dustCryotheum.name=Cryotheum Dust
 item.MiscUtils.material.rodBlizz.name=Blizz Rod
 item.MiscUtils.material.dustBlizz.name=Blizz Powder
 
-//============= 'miscutils' is Case Sensitive Here, unsure why
+//------------- 'miscutils' is Case Sensitive Here, unsure why
 tile.miscutils.fluid.pyrotheum.name=Blazing Pyrotheum
 tile.miscutils.fluid.cryotheum.name=Gelid Cryotheum
 tile.miscutils.fluid.ender.name=Resonant Ender
-//==============
+//--------------
 
 //Forestry Stuff
 item.frameAccelerated.name=Accelerated Frame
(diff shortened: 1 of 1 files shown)


draknyte1 (commented):
I don’t see what you’re doing here beyond commenting out some keys.. the keys are case sensitive, so some duplicates exist value wise, but have unique keys case wise.

MuXiu1997:
> I don’t see what you’re doing here beyond commenting out some keys.. the keys are case sensitive, so some duplicates exist value wise, but have unique keys case wise.

For example:
https://github.com/GTNewHorizons/GTplusplus/blob/3ead7cae28a1ba038345c09af14e3c3c18f3f259/src/main/resources/assets/miscutils/lang/en_US.lang#L3112-L3118

https://github.com/GTNewHorizons/GTplusplus/blob/3ead7cae28a1ba038345c09af14e3c3c18f3f259/src/main/resources/assets/miscutils/lang/en_US.lang#L3141-L3147

MuXiu1997:
Not sure what value you really want to use, but apparently the key is duplicated and an override will occur

src/main/resources/assets/miscutils/lang/en_US.lang
```diff
@@ -55,11 +55,11 @@ item.MiscUtils.material.dustCryotheum.name=Cryotheum Dust
 item.MiscUtils.material.rodBlizz.name=Blizz Rod
 item.MiscUtils.material.dustBlizz.name=Blizz Powder
 
-//============= 'miscutils' is Case Sensitive Here, unsure why
+//------------- 'miscutils' is Case Sensitive Here, unsure why
```
MuXiu1997, line 58:
```java
public static HashMap<String,String> parseLangFile(InputStream inputstream)
    {
        HashMap<String,String> table = Maps.newHashMap();
        try
        {
            Iterator iterator = IOUtils.readLines(inputstream, Charsets.UTF_8).iterator();

            while (iterator.hasNext())
            {
                String s = (String)iterator.next();

                if (!s.isEmpty() && s.charAt(0) != 35)
                {
                    String[] astring = (String[])Iterables.toArray(equalSignSplitter.split(s), String.class);

                    if (astring != null && astring.length == 2)
                    {
                        String s1 = astring[0];
                        String s2 = numericVariablePattern.matcher(astring[1]).replaceAll("%$1s");
                        table.put(s1, s2);
                    }
                }
            }

        }
        catch (Exception ioexception)
        {
            ;
        }
        return table;
    }
```
`//` is not a comment symbol

Dream-Master (approved):

PR #196, opened 2022-05-30
Add UHV Soldering Alloy for Endgame Recipes

Steelux8:
- Adder recipe for Mutated Living Solder, an endgame soldering alloy meant for the bioware circuit line and content after it;
- Added a new, very expensive catalyst for this recipe, which favors the usage of Awakened Draconium Coils to prevent losing durability on the catalyst;
- Changed the Chemical Plant so that using Awakened Draconium Coils with Tungstensteel Pipe Casings removes the chance of consuming catalyst durability.

As requested by several lategame players after adding the Indium solder, I added in this endgame one so that existing automation doesn't need to break twice. This is an UHV craft, post t8 rocket, which uses expensive materials like Bio Cells, Infinity, Gravi Stars, Plasmas and a lot of power. The recipe time is 800 seconds, but with Awakened Draconium Coils this becomes 144 seconds.

Those coils are important, because they prevent the catalyst from being consumed, an addition I made in this PR as well. The catalyst needs 1 Infinity Dust per single catalyst, which is a big incentive for these coils to be used. The Chemical Plant tier required here is Botmium, which has never been needed for any recipe, and which requires Advanced Nitinol. After talking with several lategame players, we agreed that this is a manageable expense. Additionally, Colen said that they wanted the Cyclotron to be used more, which is the multi that crafts what is needed for Advanced Nitinol.

This solder is meant for the bioware circuit line, and the recipes that use solder and come after the UEV Bio Mainframe.


Sphyix:
So 144 seconds at UHV for how much of it?
Why does everything need to be machine-spam bound or time gated...

EDIT-
Cost of it is ridicoulus aswell
64 biocells is 4 buckets of sterilized bio catalyst
it's 8x the current cost for 1 uev circuits just for soldering alloy?

boubou19:
yup this is ridiculous. If this soldering alloy is used in the same quantities as regular one in recipes, then you basically overnerfed UHV+ for the sake of it.

ScriptedPiky:
My additional problem with this is that it causes a gigantic shopping list required; you need the Cyclotron, but the recipes currently suck ass and take forever to get the process going for your single Quantum Anomaly, and are extremely RNG based. Then, you need large quantities of Botmium that also takes forever to process at UHV. And finally, it's even more expensive by the usage of any biocells, which already have intensive processing steps required and are rather annoying to make & are backlogged by some other processes.

I like the overall ideas of it, but it's so much of a shopping spree and gigantic time gate simulator, akin to smelting neutronium.

Sphyix:
Maybe this can be simplified by just using 2 or 3 metals? A gt++ alloy can be added but one with a low processing time, or better with mixer recipe instead of only ABS
I like the idea of the new soldering alloys, but we already have a bottleneck with the bacterial vats for uev, and another one isn't needed
A recipe time of 800 seconds is completely useless as well, no need to timegate a soldering alloy

GlodBlock:
i have told him to add a mixer recipe for these soldering alloy, but he is stubborn and insist on the abs. 😃

boubou19:
a mixer recipe is mandatory, or i'll fight to death against this PR.

Sphyix:
Don't throw this PR out tho, the whole thing could be used as a fluid for umv or higher circuits, maybe a bit simplified since the 1% chance after a 5%x3 chance after a 1%x9 chance to get the item for the advanced nitinol on the cyclotron is a bit ridicoulus atm

For soldering alloy, this is definitely not possible

chochem:
I dont know what you guys mean by a mixer recipe. Replacing a complex chem plant recipe with catalyst by just a mixer is not even possible and not a reasonable suggestion. Are you mistaking this for the other soldering alloy? 

I really like the proposal overall. Steelux, maybe you can chat a bit more with a few UHV+ players as to what would be a reasonable processing time (at the moment this looks like 2 ingots worth per second at UHV) and resource amounts (bio cells in particular). Its been too long for me to have a clear idea. Of course this also heavily depends on how much of this alloy is actually needed in recipes. If they just need a small trickle it might be fine as is. (edit: steelux clarified it would probably replace the old one 1:1)

boubou19:
> I dont know what you guys mean by a mixer recipe. Replacing a complex chem plant recipe with catalyst by just a mixer is not even possible and not a reasonable suggestion. Are you mistaking this for the other soldering alloy?

The abs recipes have an alternative with mixer + ebf recipes. It's way more scalable than ABS. Bottlenecks shouldn't be a thing UHV+.

joshaup1999:
Mixer recipe is for the ABS alloy, it’s a stupidly high time gate just for the sake of it

chochem:
As I pointed out boubou, this is not an ABS recipe.

GlodBlock:
you need botmium and some other high voltage abs alloy to build chem planet. also the previous alloy he added in assline is made in abs.

src/main/java/gtPlusPlus/core/item/chemistry/GenericChem.java
```diff
@@ -484,6 +491,34 @@ private void recipeCarbonDisulfide() {
...
+						FluidUtils.getFluidStack("plasma.bismuth", 18000),
+						FluidUtils.getFluidStack("cryotheum", 4000)
+				},
+				new ItemStack[] {
+
+				},
+				new FluidStack[] {
+						MISC_MATERIALS.MUTATED_LIVING_SOLDER.getFluidStack(40000)
+				},
+				20*800,
```
boubou19, line 516: 800s at UEV for 40buckets, that's like 2.5mb/t if i didn't made any miscalculation.
I'd keep that ratio, but would divide the time by 4.

GTNHAfx237v7:
But that was 100 ingots in 40 seconds (IV?) so pretty fast anyway

src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/chemplant/GregtechMTE_ChemicalPlant.java
```diff
@@ -935,7 +935,11 @@ private ItemStack findCatalyst(ItemStack[] aItemInputs, ItemStack[] aRecipeInput
 
 	private void damageCatalyst(ItemStack aStack, int parallelRecipes) {
 		for (int i=0; i<parallelRecipes; i++){
-			if (MathUtils.randFloat(0, 10000000)/10000000f < (1.2f - (0.2 * this.mPipeCasingTier))) {
+			// Awakened Draconium Coils with Tungstensteel Pipe Casings (or above) no longer consume catalysts.
+			if (this.mCoilTier >= 11 && this.mPipeCasingTier >= 4) {
```
boubou19, line 939: is `this.mCoilTier` the meta of the coil blocks or its actual tier? because if it's the meta then they aren't all sorted.

GlodBlock:
abs only supports one energy hatch, so you can't overclock it until you made the first luv hatch. also it doesn't have any parallel

src/main/java/gtPlusPlus/core/item/chemistry/GenericChem.java
Steelux8: It's 144 seconds per recipe, I mentioned this on the PR itself.

src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/chemplant/GregtechMTE_ChemicalPlant.java
Steelux8: There is a switch case elsewhere in the Chemical Plant code that translates the unsorted coil tiers to an ordered list of numbers, which is what's used here.

src/main/java/gtPlusPlus/core/item/chemistry/GenericChem.java
Sphyix: I expect to use less than 50mb of soldering for every circuit if that's the case

Sphyix:
> I dont know what you guys mean by a mixer recipe. Replacing a complex chem plant recipe with catalyst by just a mixer is not even possible

That's why this chemline is fine, but not for soldering alloy
Use this for some other fluids for high tier circuits, and make soldering alloy something expensive but that doesn't require a stupid amount of time to make. Uhv+ circuits are already time gated (or require massive machine spam). We don't need more timegates

Steelux8:
> > I dont know what you guys mean by a mixer recipe. Replacing a complex chem plant recipe with catalyst by just a mixer is not even possible
> 
> That's why this chemline is fine, but not for soldering alloy Use this for some other fluids for high tier circuits, and make soldering alloy something expensive but that doesn't require a stupid amount of time to make. Uhv+ circuits are already time gated (or require massive machine spam). We don't need more timegates

What are UEV Bio Mainframes used for that is not "Endgame Goals" endgame? Those mainframes are supposed to be really difficult to make, and the UHV players I spoke with to decide on the materials and amounts didn't disagree with that.

src/main/java/gtPlusPlus/core/item/chemistry/GenericChem.java
Steelux8: Please explain your reasoning for why circuits would need such a small amount with this solder.

Sphyix:
> > > I dont know what you guys mean by a mixer recipe. Replacing a complex chem plant recipe with catalyst by just a mixer is not even possible
> > 
> > 
> > That's why this chemline is fine, but not for soldering alloy Use this for some other fluids for high tier circuits, and make soldering alloy something expensive but that doesn't require a stupid amount of time to make. Uhv+ circuits are already time gated (or require massive machine spam). We don't need more timegates
> 
> What are UEV Bio Mainframes used for that is not "Endgame Goals" endgame? Those mainframes are supposed to be really difficult to make, and the UHV players I spoke with to decide on the materials and amounts didn't disagree with that.

All the circuits above uev are already on dev, it can be used for that.
What are you guys going to do for umv or UXV circuits ?
1 day processing time and a 0.1% chance of crafting it?

Use this fluid for uv circuit line, so uv, uhv, uev, uiv (current uiv is nano) and uhv circuit line, so uhv, uev, uiv, umv (current umv is piko)

Don't make existing content more tedious by requiring longer and longer processing times and more machines, use those for the new content that was partially already added.
All the uv+ lines circuits up to uev line (uev, uiv, umv, UXV) are already in dev i think, they just require a crafting recipe

Steelux8:
I'm not responsible for those new circuits. What are they meant to be used for? What is the crafting recipe supposed to be like? The main goal of the bioware line is to make UEV circuits for the endgame, so I tiered this solder recipe accordingly. While doing that, I grabbed two multis that didn't really have a necessary use, the Botmium Chemical Plant and the Cyclotron, and they are now used here.

For UHV players, of course, it's easy to just pump more energy into paralleled Cyclotrons to accelerate the Quantum Anomaly. That is what I got as feedback. Is it actually a timegate for such a lategame tier?

src/main/java/gtPlusPlus/core/item/chemistry/GenericChem.java
boubou19: with the awakened draconium coils it becomes 13.88 mb/t. depending on the usage of such new alloy, it may become a bottleneck again, unless you use low amount in recipes as @Sphyix said: otherwise you'll need to spam machines again, and that's boring. (Also the power consumption dedicated to this lol)

Steelux8: I'll keep the recipe time and lower the energy usage by one tier, then. Should be good enough then, with the added parallel.

Sphyix:
Not asking you to be responsible for those circuits, just saying that something like what you did here makes a lot more sense on the new lines instead of making existing lines more complicated.
Cyclotron doesn't have parallels btw

Sphyix:
Anyway, just so you get my point, i'm totally not against the new soldering alloys, they need to exist since what is being done now doesn't make sense, same soldering lv to uev.
I'm just against nerfing stuff for the sake of it, new soldering doesn't need to be over complicated or doesn't need to require crazy amounts of times to make, it's just a soldering alloy, make it expensive and fast to process.

I understand that "expensive" at uev is not really a thing since we have void miners, but making it require crazy amounts of time is just going to push players to spam more machines.

I mean, just as an example, this is my neutronium farm
javaw_43sORSBHCc

Neutronium is a time gate, overclocking is pretty much impossible, so what you see on the screenshot is what happens.
This is not a fun mechanic, spamming the same machine, in my case, 128 times.

I don't want that to happen on UEV circuits


-EDIT- (kinda off topic, I guess)
Just to explain my point on the new circuit lines.
Right now we have zpm circuit line that does zpm, uv, uhv, uev, uiv (nano), umv (piko), uxv (quantum)
That's a 7 time recursion, getting recipes for a uv circuit line or a uhv circuit line would be nice, instead of making current zpm line more complex, that's why suggestion a few posts above was to use this new fluid for a new line, like the bio fluid is used now on the bio circuits.
It's complicated enough to make sense for a new circuit

-EDIT 2- 
To reply to your questions about the new circuits aswell.
Their recipe doesn't exist. What it needs to be like can be decided by whoever makes it, I guess, like all the other recipes were decided.
Their use is the current use of nano, piko and quantum circuits, at least for now.

Steelux8:
From your logic, then, why not apply this solder to the bioware circuit line and the new circuits after that one? The amount of solder needed for the biowares can be smaller than the current solder amount, that's perfectly fair.

Sphyix:
> From your logic, then, why not apply this solder to the bioware circuit line and the new circuits after that one? The amount of solder needed for the biowares can be smaller than the current solder amount, that's perfectly fair.

This fluid that you made, in my opinion, is too complicated and requires too much time to make to be used as a soldering alloy, unless it is used in waay smaller quantities.

An alternative can be to keep the overall cost for uev the same, so more expensive soldering alloy and less expensive circuit, then it's a fair trade. Otherwise it's a nerf for uev for the sake of nerfing it.

Let's say quantity for soldering alloy is kept the same.
40 ingots of new soldering (5760L) is 64 biocells.
Soldering circuit cost is:
72L for 16 ZPM (let's not count them)
144L for 16 UV (let's not count them either)
2880L for 2 UHV (32 biocells)
2880L for 1 UEV (32 biocells)

Current cost for 1 UEV is 128 biocells (8 zpm circuits, 16 biocells each) 
Let's reduce the cost for bio processing units by 8, so it translates to 8 biocells for each zpm circuit.
This way the extra 8 biocells can be used for the soldering.

Time for the soldering MUST be reduced to something more acceptable, 5760L every 144 seconds is a massive timegate, it's less than 1 circuit every 144 seconds.
I'd say something more acceptable can be 1 circuit every 10 sec? Voltage could even be fine but time is not.

tl;dr
Idea can be to:
Reduce biocells quantity in the craft of Bio Processing Unit by 8, and use those for soldering, keeping the overall cost of UEV the same.
Reduce the crafting time of the soldering to be able to do 1 circuit every 10 seconds.

Steelux8:
> Let's say quantity for soldering alloy is kept the same.
40 ingots of new soldering (5760L) is 64 biocells.

I'm sorry, but each recipe consumes 64 biocells to make 40 buckets (40000L), not 40 ingots.

Sphyix:
> > Let's say quantity for soldering alloy is kept the same.
> > 40 ingots of new soldering (5760L) is 64 biocells.
> 
> I'm sorry, but each recipe consumes 64 biocells to make 40 buckets (40000L), not 40 ingots.

Oh well then ignore my calculations.
I definitely can't read LOL

Steelux8:
So, are there any further wanted changes? Should I change anything at all?

Sphyix:
> So, are there any further wanted changes? Should I change anything at all?

Yes, this:

> I'll keep the recipe time and lower the energy usage by one tier, then. Should be good enough then, with the added parallel.

ScriptedPiky:
> So, are there any further wanted changes? Should I change anything at all?

I would suggest adjusting the RNG and the time on the Cyclotrons if it's possible. It could be at ZPM tier but lasts 1 second or less, and is a little more forgiving on RNG instead of the 1-5% hell

Sphyix:
Cyclotron fix could even be parallels or perfect OC with same chances

Steelux8:
Cyclotron recipes were buffed, I think this UHV solder is mostly good to go now?

Dream-Master (approved):

PR #201, opened 2022-06-04
Add Indalloy_140 mixer/EBF recipe

Elisis:


chochem:
adding mixer recipes to alloys sounds pretty boring and simplifying. Is that really needed?

GTNHAfx237v7:
Seconded, it’s made in the ABS at a reasonable speed already.

Connor-Colenso (changes requested):
Unnecessary. ABS is sufficient as others have noted.

Steelux8:
Do not set "uses Blast Furnace" to "true" on Indalloy, that is for auto-generating the recipe and it will have different recipe times based on whether or not CoreMod is loaded. I've fixed this already.

Also, as mentioned before, mixer recipe is unnecessary and an oversimplification. At the very least, it should be much slower than the ABS recipe and a higher tier.

Elisis:
> Unnecessary. ABS is sufficient as others have noted.

And others have argued against 🙄

chochem:
> > Unnecessary. ABS is sufficient as others have noted.
> 
> And others have argued against 🙄

what is your argument then? that was my question above.

Dream-Master:
set it to draft until we find a way all people happy

draknyte1:
Unnecessary. ABS is sufficient as others have noted.

Stop trying to remove the use of an entire mod. It’s gated behind the ABS, it should stay that way. It’s like me adding recipes to circumvent the CAL or some other multi with one job. This should be closed.

PR #203, opened 2022-06-05
More saner & less tedious recipes

ScriptedPiky:
**Time in ticks**
Quantum Anomaly: 1000 > 750
Hydrogen Ion Production: 15000 > 2500
Hydrogen Plasma: 1 > 100
Tier is LuV; unchanged (or is it IV?)

### DIFF
--- a/src/main/java/gtPlusPlus/core/recipe/RECIPES_GREGTECH.java
+++ b/src/main/java/gtPlusPlus/core/recipe/RECIPES_GREGTECH.java
@@ -2192,10 +2192,10 @@ public class RECIPES_GREGTECH {
 						GregtechItemList.Laser_Lens_Special.get(1)
 				}, 
 				null,
-				new int[] {1500},
+				new int[] {100},
 				20 * (GTNH ? 300 : 60),
-				(int) MaterialUtils.getVoltageForTier(7),
-				750 * 20);
+				(int) MaterialUtils.getVoltageForTier(8),
+				500 * 20);
 
 		/*
 		 * Ions
--- a/src/main/java/gtPlusPlus/core/recipe/RECIPES_GREGTECH.java
+++ b/src/main/java/gtPlusPlus/core/recipe/RECIPES_GREGTECH.java
@@ -2195,7 +2195,7 @@ public class RECIPES_GREGTECH {
 				new int[] {100},
 				20 * (GTNH ? 300 : 60),
 				(int) MaterialUtils.getVoltageForTier(6),
-				1000 * 20);
+				750 * 20);
 
 		/*
 		 * Ions
@@ -2266,7 +2266,7 @@ public class RECIPES_GREGTECH {
 				new int[] { 125, 125, 125, 125, 125, 125, 125, 125, 125 },
 				20 * 20,
 				(int) MaterialUtils.getVoltageForTier(6),
-				15000);
+				2500);
 
 		// Generate Hydrogen Plasma Recipe
 		CORE.RA.addCyclotronRecipe(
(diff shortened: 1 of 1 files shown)


miozune:
Can we also increase chances (and duration as well), especially for Quantum Anomaly? 1% is too small.
Yes IRL particle collisions are RNG, but it's bad for game design. Lucky people can cheese it, while unlucky people are pissed off.

ScriptedPiky:
Does increasing chance to 25% and voltage tier to 7 (I think ZPM?) work? And duration in ticks? Would 500 ticks work? Or rather it's currently 750 ticks

draknyte1:
> Can we also increase chances (and duration as well), especially for Quantum Anomaly? 1% is too small.
> Yes IRL particle collisions are RNG, but it's bad for game design. Lucky people can cheese it, while unlucky people are pissed off.

Every player I’ve seen make quantum anomalies didn’t complain, some even felt it rewarding when it finally appeared in their output bus. End game players who need several didn’t find it too costly or time consuming.

Eranicus:
stop making endgame easyer please....  anomaly's were never a problem, its called anomaly for a reason and should be as rare as 1%

ScriptedPiky:
I myself found it incredibly tedious, mostly since it's a 1-off kind of thing but it's so absurdly rare that it's more or less a time gate; specifically it took over a day of 4 UV paralleled Cyclotrons to get the required low amount of Anomalies (around 3-4 at the time). I could potentially make it 1% again but increasing it to UV or UHV with a somewhat short processing time

Eranicus:
As a UHV player, I managed to build 32 mass fabricators which require these anomalies in 3-4 days time.... without any problems with 3 cyclotrons. In my opinion the chance should remain the same (1%), processing time can be decrised if you want but still not neccesary in my opinion... end game should be hard

Steelux8:
Anomalies are supposed to be rare, and you should be able to build several Cyclotrons to get them for endgame purposes. It's unreasonable to expect it to come easily off one Cyclotron. The 1% chance is fine as it is.

miozune:
I mean, I don't mind the total energy usage, what I'm worried about is that low chance... but I'm still at mid-ZPM and never built one, so if most of the people say fine as is, I'll follow the pioneers.

ScriptedPiky:
> Anomalies are supposed to be rare, and you should be able to build several Cyclotrons to get them for endgame purposes. It's unreasonable to expect it to come easily off one Cyclotron. The 1% chance is fine as it is.

It's now returned to 1% and is UV tier at 25 seconds.

> I mean, I don't mind the total energy usage, what I'm worried about is that low chance... but I'm still at mid-ZPM and never built one, so if most of the people say fine as is, I'll follow the pioneers.

The most tedious part was the hydrogen ion production, as it took 15000 ticks or rather, 750 seconds at LuV, and the chances were 1.25% for each, which is what caused the enormous bottleneck to begin with. Technically, you can only use the Quantum Anomaly at UV tier-ish, though I'm not sure if 2 or more energy hatches works on the Cyclotron

Eranicus:
> I'm not sure if 2 or more energy hatches works on the Cyclotron

They do work, i used x2 hatches per cyclotron in order to decrease prod. time while I was farming anomalies.

ScriptedPiky:
Oh perfect! That's excellent then.

Steelux8:
This change would also solve the last problem with the endgame solder in #196.

Steelux8 (approved):
Reduces the time gate on the Quantum Anomaly farm if enough energy is put in, looks good to me.

PR #204, opened 2022-06-06
Buff Sphalerite milling

miozune:
aluminium way : 9x Crushed Sphalerite Ore -> 1x Indium
milling (before) : 12.8 -> 1
milling (after) : 8 -> 1

Milling machines should fill the gap between EV-tier dumb aluminium way and UHV-tier T3 fusion.

I also adjusted milling duration for plain ore and crushed ore to be the same, since already crushed ore taking longer time doesn't make sense.


GlodBlock (approved):
finally it's cheaper than replicating with uum 😋

PR #207, opened 2022-06-09
Boost XL Turbine Speed and Allow TecTech Hatches

Steelux8:
- Added the option to use Tectech Dynamo Hatches on XL turbines;
- Tweaked the current speed from 18x to the 16x that Alkalus told me was intended (not 100% sure about this, breaks current setups somewhat but my testing gave a higher value than intended);
- Added Fast Mode, which can be switched into to triple the multi's speed to 48x, but with greatly accelerated (12x) turbine damage and maintenance problems;
- Integrated Loose Mode into the mode changing for Steam and SH Steam XL Turbines, so that any combination is possible;
- Added information to the tooltip to clarify the numbers and speed.

**All numbers can be adjusted, feedback is welcome.**

While looking at the XL Turbines, I realized that they only run a bit faster than the same volume of regular Large Turbines. One XL takes the space of 12 Large, but my testing showed a 18x speed difference, which is 50% higher per volume. Of course, having all the inputs and outputs + dynamos in the same multi is fantastic, but all mega multis give a much faster speed boost than this. Upon reaching endgame powergen that uses turbines, like fusion, spamming XL turbines will be required at a certain point, which I think should not be intended.

To not force everyone into 48x speed, I created two modes and made Slow Mode the default. The Steam Turbines' Loose mode can be interchanged with these two in any desired combination. Optimal flow and pollution raise in the same right, but problems such as maintenance and turbine damage become much more frequent in Fast Mode.

This huge speed boost merits a recipe change to the XL Turbines to make them more expensive, in my opinion. On that end, I'd like to hear suggestions.

### DIFF
--- a/src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/turbines/GT_MTE_LargeTurbine_Plasma.java
+++ b/src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/turbines/GT_MTE_LargeTurbine_Plasma.java
@@ -107,7 +107,7 @@ public class GT_MTE_LargeTurbine_Plasma extends GregtechMetaTileEntity_LargerTur
 					float aTotalBaseEff = 0;
 					float aTotalOptimalFlow = 0;
 					ItemStack aStack = getFullTurbineAssemblies().get(0).getTurbine();
-					for (int i=0;i<18;i++) {
+					for (int i=0;i<speedMultiplier;i++) {
 						if (i == 0) {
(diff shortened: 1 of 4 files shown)


draknyte1 (changes requested):
TT stuff is already handled internally. You can just pass anything extending Hatch to addToMachineList. Checking for TT here just results in it being done again in the internals. 

All dynamos are added to a mixed map which can then be accessed in the event people mix. I believe there’s also a method which can be called to clear all 3 Dynamo lists.

Steelux8:
What do you mean by "handled internally"? TT dynamos didn't work from my testing, and then they worked but a regular dynamo was required, and now the TT dynamo works by itself.

Dream-Master:
@draknyte1

Dream-Master (approved):

PR #213, opened 2022-06-13
Algae Farm Buffs and Tweaks

Steelux8:
- Changed the Algae Farm cycles, so that the multi works much faster but outputs less algae, in a more similar fashion to other multiblocks;
- Tweaked some related recipes that were still too underpowered;
- Increased the cost of Compost per tier to reinforce the decision between using it, upgrading the Algae Farm or making more.

This PR is (hopefully) the final change to make the Algae Farm usable and more popular within progression. This multi runs on no power to generate algae, slowly, but running faster as its tier increases. From those algae, the player can automate many different kinds of items and fluids, including chemicals that are essential to progression like Sulfuric Acid, Ethylene, Methane, etc.

The recipe paths to obtain those and other outputs are generally better than the conventional ways, but at the cost of having to build Chemical Plants, since they are required for almost all recipes in the Algae Farm processing line. However, this structure investment should pay off with the automation of several important things at once, from a single source, and even some power generation from Methane or, later, Butanol.

### DIFF
--- a/src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/algae/GregtechMTE_AlgaePondBase.java
+++ b/src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/algae/GregtechMTE_AlgaePondBase.java
@@ -49,6 +49,7 @@ public class GregtechMTE_AlgaePondBase extends GregtechMeta_MultiBlockBase<Gregt
 	private int mCasing;
 	private IStructureDefinition<GregtechMTE_AlgaePondBase> STRUCTURE_DEFINITION = null;
 	private int checkMeta;
+	private int minTierOfHatch = 100;
 	private static final Class<?> cofhWater;
 
 	static {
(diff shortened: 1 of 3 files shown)


chochem:
Glad to see the algae farm seeing some love. the old recipes were just really badly balanced for GTNH. between the last change by steelux and this one it should now be an interesting multi that also brings some more engagement to the chem plant. :)

I havent not checked the precise numbers, will test that once its in. Using i/o hatches for tiering is certainly a good idea (it was mine :P) as there are no energy hatches and some high tier casings are easily to skip.

GTNHAfx237v7:
What tiers of casings does this now allow?

Steelux8:
> What tiers of casings does this now allow?

It should allow all of them, from what I know.

GTNHAfx237v7:
Cool!

src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/algae/GregtechMTE_AlgaePondBase.java
```diff
@@ -82,6 +83,8 @@ protected GT_Multiblock_Tooltip_Builder createTooltip() {
 				.addInfo("Provide compost to boost production by one tier")
 				.addInfo("Does not require power or maintenance")
 				.addInfo("All Machine Casings must be the same tier, this dictates machine speed.")
+				.addInfo("All Buses/Hatches must, at least, match the tier of the Casings")
+				.addInfo("Place the controller after everything else to make sure it forms correctly!")
```
eigenraven: I'm pretty sure if you add a pass with `minTierOfHatch = 100;` at the beginning of the structure definition/validation code, the placement order won't matter

Steelux8: Oh, is that why the controller sometimes had to be placed last? I never figured it out.

Steelux8: Okay, I fixed it and it recognizes the right hatches properly now.

eigenraven (approved):

PR #216, opened 2022-06-15
Fix recipe inconsistency with UHV Battery

miozune:


Steelux8 (approved):
Thank you for fixing this on my behalf.

PR #217, opened 2022-06-16
Fix fertilizer usage in Crop Manager

xSkewer:
Haven't tested ingame yet but it should work. Also increases the fertilizer cap to 200, which is what Forestry farms are at.

### DIFF
--- a/src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/basic/GT_MetaTileEntity_CropHarvestor.java
+++ b/src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/basic/GT_MetaTileEntity_CropHarvestor.java
@@ -421,7 +421,7 @@ public class GT_MetaTileEntity_CropHarvestor extends GT_MetaTileEntity_BasicTank
 			return false;
 		} else {
 			//Logger.INFO("Current Nutrient: "+aCrop.getNutrientStorage()+" for "+aCrop.getCrop().displayName());
-			aCrop.setNutrientStorage(100);
+			aCrop.setNutrientStorage(200);
 			return true;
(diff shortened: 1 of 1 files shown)


src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/basic/GT_MetaTileEntity_CropHarvestor.java
```diff
@@ -421,7 +421,7 @@ public boolean applyFertilizer(ICropTile aCrop) {
 			return false;
 		} else {
 			//Logger.INFO("Current Nutrient: "+aCrop.getNutrientStorage()+" for "+aCrop.getCrop().displayName());
-			aCrop.setNutrientStorage(100);
+			aCrop.setNutrientStorage(getNutrientStorage() + 100);
```
kuba6000: I think you forgot something
```suggestion
			aCrop.setNutrientStorage(aCrop.getNutrientStorage() + 100);
```

kuba6000 (changes requested):

kuba6000 (approved):
Ok

draknyte1:
> Haven't tested ingame yet but it should work. Also increases the fertilizer cap to 200, which is what Forestry farms are at.

IC2 states the max should be 100. The values for the crop manager are based on IC2. 

Forestry should be nerfed, it’s essentially giving crops free fert by setting to 200.

xSkewer:
> IC2 states the max should be 100. The values for the crop manager are based on IC2.
> 
> Forestry should be nerfed, it’s essentially giving crops free fert by setting to 200.

You can hand apply fertilizer up to 199.

PR #259, opened 2022-08-01
Fix recipe check for ABS input bus separation

xSkewer:
With only the first change (https://github.com/GTNewHorizons/GTplusplus/pull/257), the ABS would check every input bus, but never check any recipes with the inputs, resulting in it only using one of its input buses.

This change runs a recipe check for each input bus until it finds a recipe or reaches the last input bus. It also changes the order that the ABS gets its inputs, fluids, and voltage tier in to make this possible.

### DIFF
--- a/src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntity_AlloyBlastSmelter.java
+++ b/src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntity_AlloyBlastSmelter.java
@@ -261,9 +261,9 @@ public class GregtechMetaTileEntity_AlloyBlastSmelter extends GregtechMeta_Multi
 							}
 						}
 					}
-					tInputs = Arrays.copyOfRange(tInputList.toArray(new ItemStack[tInputList.size()]), 0, tInputList.size());
+					tInputs = tInputList.toArray(new ItemStack[0]);
 					tRecipe = GTPP_Recipe.GTPP_Recipe_Map.sAlloyBlastSmelterRecipes.findRecipe(this.getBaseMetaTileEntity(), false, gregtech.api.enums.GT_Values.V[tTier], tFluids, tInputs);
-					if ((tRecipe != null) && (tRecipe.isRecipeInputEqual(false, tFluids, tInputs))) {
+					if ((tRecipe != null)) {
 						break;
 					} 
 				}
@@ -281,7 +281,7 @@ public class GregtechMetaTileEntity_AlloyBlastSmelter extends GregtechMeta_Multi
 						}
 					}
 				}
-				tInputs = Arrays.copyOfRange(tInputList.toArray(new ItemStack[tInputList.size()]), 0, tInputList.size());
+				tInputs = tInputList.toArray(new ItemStack[0]);
 				tRecipe = GTPP_Recipe.GTPP_Recipe_Map.sAlloyBlastSmelterRecipes.findRecipe(this.getBaseMetaTileEntity(), false, gregtech.api.enums.GT_Values.V[tTier], tFluids, tInputs);
 			}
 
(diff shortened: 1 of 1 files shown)


src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntity_AlloyBlastSmelter.java
```diff
@@ -255,6 +261,11 @@ public boolean checkRecipe(final ItemStack aStack) {
 							}
 						}
 					}
+					tInputs = Arrays.copyOfRange(tInputList.toArray(new ItemStack[tInputList.size()]), 0, tInputList.size());
+					tRecipe = GTPP_Recipe.GTPP_Recipe_Map.sAlloyBlastSmelterRecipes.findRecipe(this.getBaseMetaTileEntity(), false, gregtech.api.enums.GT_Values.V[tTier], tFluids, tInputs);
+					if ((tRecipe != null) && (tRecipe.isRecipeInputEqual(false, tFluids, tInputs))) {
```
GlodBlock: why you check the recipe again when it already find one?

src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntity_AlloyBlastSmelter.java
```diff
@@ -255,6 +261,11 @@ public boolean checkRecipe(final ItemStack aStack) {
 							}
 						}
 					}
+					tInputs = Arrays.copyOfRange(tInputList.toArray(new ItemStack[tInputList.size()]), 0, tInputList.size());
```
GlodBlock: why copy the array? i don't think it will cause problem if you directly it.

src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntity_AlloyBlastSmelter.java
```diff
@@ -270,8 +281,9 @@ public boolean checkRecipe(final ItemStack aStack) {
 						}
 					}
 				}
+				tInputs = Arrays.copyOfRange(tInputList.toArray(new ItemStack[tInputList.size()]), 0, tInputList.size());
```
GlodBlock: the same here

GlodBlock (changes requested):

Dream-Master:
@xSkewer please recheck your pr. <s>Cant reach you on Discord anymore. Pm is not possible</s>

src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntity_AlloyBlastSmelter.java
xSkewer: This was how the original code did it. I just moved the lines to the input bus checks.

src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntity_AlloyBlastSmelter.java
xSkewer: This was how the original code did it.

src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntity_AlloyBlastSmelter.java
xSkewer: Same as above.

src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntity_AlloyBlastSmelter.java
GlodBlock: alk's crap code probably, could you just remove them?

GlodBlock (approved):

PR #270, opened 2022-08-05
Lock multiblock bonuses behind tiered covers

alexandre-tardif:
This feature locks the bonuses of every multis that are flagged as a tiered multi. This makes multi cost a little more than an energy hatch.
That way while people bases will become slower it won't brick any bases as all the autocrafting will still work but be slower until you make the covers.
It still allows overclocking with two hatches but heavily discourages overclocking tiered machines over your current tier
[WIP]


chochem:
overall I like the approach but I certainly have some questions with this one. Maybe can you add some more explanations.

Like:
1. why a cover? Correct me if I am wrong but I always thought they are pretty tps hungry, isnt there a better way to implement this?
2. is the cover you need based on recipe tier or energy tier? Imo it should _definitely_ be recipe tier, you shouldn't need a higher tier cover to do parallels or 4/2 OCs. You can also feed lots of amps into a PA for parallels without needing any higher tier components.

alexandre-tardif:
1. The biggest reason to go with a cover was to not need any extra hatches/buses or having to use an item slot for it. I can look at if it has some performance issue but right now the only thing it does is set a value when placed and removed
2. Right now it's energy tier. I can look at doing recipe tier. Still a WIP

miozune:
I'd say hold on until Mitch finishes MTE. It will allow modularity in multiblock structure.

alexandre-tardif:
From what I understand, using mitch MTE would require a full rework of every GTPP multi. This PR would add cost for the current GTPP multi without breaking bases

chochem:
1. tbf I am not too sure on that. I guess if it doesn't tick its all good. :)
2. oh I totally get its WIP. ofc you can/should ask what other people say, but as argued above I definitely think it should be recipe tier.

hmm, I dont think Mitch MTEs _really_ plays into this. I mean we wont remove all the old multis (at least not immediately!) just because we have the possibility for new ones.

boubou19:
There is a disabled but similar gate in the form of the control cores. What's the advantage of such new covers in comparison to a possible rework of the control cores?

alexandre-tardif:
> There is a disabled but similar gate in the form of the control cores. What's the advantage of such new covers in comparison to a possible rework of the control cores?


From what I understand control core made it so you're unable to OC past the core and needed a specific bus for them. I didn't really like the implementation because 
1. It would break bases
2. It requires a bus. While we can reduce minimum casing for each multi that would (while not much) reduce the cost of all multis when you first make them

Dream-Master:
@alexandre-tardif reopen if you want to finish

PR #271, opened 2022-08-05
Flotation Cell fixes

### LINKED ISSUE #10924: GT++ Flotation Cell Regulator locked to invalid value again

GregTechDriver

### Your Pack Version

2.1.2.4-25

### Your Server

SP

### Type of Server

Single Player

### Your Expectation

Just want to process some indium from sphalerite

### The Reality

Doesn't work
The ore key locked to -1590074919
Can't process any Ore

### Your Proposal

fix

### Final Checklist

miozune:
- Use material unlocalized name instead of Alk's hash
  - I'm shocked at what Alk used for hash: https://github.com/GTNewHorizons/GTplusplus/blob/master/src/main/java/gtPlusPlus/core/util/data/AES.java This is terrible. Also showing hash is user-hostile. So I changed to use material name instead.
- Fix `findMilledStack` is called with incorrect argument `aRecipe`
- Fix `findMilledStack` not working
  - This is probably why the hash is locked to value that no ore has. Still not sure why it worked with my env though.
  - Close https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/10924

### DIFF
--- a/src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMTE_FrothFlotationCell.java
+++ b/src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMTE_FrothFlotationCell.java
@@ -291,14 +291,6 @@ public class GregtechMTE_FrothFlotationCell extends GregtechMeta_MultiBlockBase<
 			foundMaterialName = foundMaterial.getUnlocalizedName();
 		}
 
-		// Set material locked for this controller
-		if (lockedMaterialName == null) {
-			if (foundMaterialName != null) {
-				lockedMaterialName = foundMaterialName;
-			}
-		}
-
-		// Check material match
 		if (foundMaterialName == null) {
 			log("Did not find material from Milled Ore.");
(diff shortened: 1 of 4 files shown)


src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMTE_FrothFlotationCell.java
```diff
@@ -275,26 +279,38 @@ public boolean checkRecipeGeneric(
...
+
+		// Set material locked for this controller
+		if (lockedMaterialName == null) {
+			if (foundMaterialName != null) {
+				lockedMaterialName = foundMaterialName;
+			}
+		}
+
+		// Check material match
+		if (foundMaterialName == null) {
```
xchgeax, line 294: This can probably be moved above `lockedMaterialName` null check and nested `foundMaterialName` null check then can be omitted.

draknyte1:
It was quick and consistent, least compared to anything else. I believe that’s why I did that, but who knows. 🤷🏻‍♀️

That class was originally used for secret things, hence the escaped Unicode string instead of a class.

xchgeax (approved):

PR #273, opened 2022-08-10
Disable the large generator array

alexandre-tardif:
The multi doesn't work as intended and allows for free power generation. This PR makes them uncraftable and makes them stop working. Code is still there if someone wants to work on them at some point

### DIFF
--- a/src/main/java/gtPlusPlus/core/recipe/RECIPES_Machines.java
+++ b/src/main/java/gtPlusPlus/core/recipe/RECIPES_Machines.java
@@ -263,7 +263,6 @@ public class RECIPES_Machines {
 		gt4Redstone();
 		gt4Inventory();
 		
-		multiGeneratorArray();
 		multiForgeHammer();
 		multiMolecularTransformer();
 		multiXlTurbines();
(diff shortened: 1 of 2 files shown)


xSkewer (approved):

xSkewer:
I normally don't approve PRs but we were looking at this multi in discord and it behaves very differently from how the tooltip describes and is also overpowered because of the fuel consumption bug. It should be disabled as soon as possible and should stay disabled until it is fixed.

Dream-Master (changes requested):
Please comment code out so we can see original code. For better fixing @alexandre-tardif

alexandre-tardif:
The code is commented. I just did a comment block so it doesn't show properly on Github. Unless you mean the recipe? I can readd it and comment it if that's what's wanted

Dream-Master:
yes the recipe. So keep the code all of it just comment it out (these parts that disable recipe etc) @alexandre-tardif

alexandre-tardif:
Done

Dream-Master (approved):

PR #278, opened 2022-08-15
Update industrial sledgehammer and forge casing recipes to be easier

BlueTRaven:
Solves https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/11024

Updated the Industrial Sledgehammer recipe to be more in line with other multiblocks from its tier. As mentioned in the issue, the industrial sledgehammer is, if not _expensive,_ more complicated to craft than most other multiblocks on its tier. I aim to bring it and its casings (which suffer from the same problem) more in line with this pull request.

Updated controller recipe:
image
Swapped the integral encasement IV to be 2 IV machine hulls. It is otherwise unchanged.

 Updated casing recipe:
image
Completely redone in a crafting table recipe instead. Requires 4 incoloy-ma956 plates, 1 babbit alloy plate, 3 hastealloy-X rods, and 1 heat proof machine casing.

### DIFF
--- a/src/main/java/gtPlusPlus/core/recipe/RECIPES_Machines.java
+++ b/src/main/java/gtPlusPlus/core/recipe/RECIPES_Machines.java
@@ -483,34 +483,28 @@ public class RECIPES_Machines {
 	}
 
 	private static void multiForgeHammer() {
-		
+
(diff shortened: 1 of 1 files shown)


Runakai1 (approved):
lgtm

PR #280, opened 2022-08-17
Autoplace integration

Glease:
Depends on https://github.com/GTNewHorizons/GT5-Unofficial/pull/1245

### DIFF
--- a/src/main/java/gtPlusPlus/xmod/gregtech/api/metatileentity/implementations/base/GT_MetaTileEntity_Hatch_CustomFluidBase.java
+++ b/src/main/java/gtPlusPlus/xmod/gregtech/api/metatileentity/implementations/base/GT_MetaTileEntity_Hatch_CustomFluidBase.java
@@ -1,44 +1,48 @@
 package gtPlusPlus.xmod.gregtech.api.metatileentity.implementations.base;
 
+import gregtech.GT_Mod;
+import gregtech.api.interfaces.ITexture;
+import gregtech.api.interfaces.tileentity.IGregTechTileEntity;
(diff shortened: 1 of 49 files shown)


src/main/java/gtPlusPlus/preloader/asm/transformers/ClassTransformer_GT_Client.java
```diff
@@ -81,6 +82,10 @@ public ClassTransformer_GT_Client(byte[] basicClass, boolean obfuscated) {
 		FMLRelaunchLog.log("[GT++ ASM] Gregtech Client Proxy Patch", Level.INFO, "Fixing Issues in vanilla GT.");		
 		if (mModern) {
 			aTempReader.accept(new MethodAdaptor2(aTempWriter), 0);
+			if (mGTNH) {
```
draknyte1: I don’t think there’s any point doing this anymore as it’s certainly only compatible with GTNH after the last 6mo of reworks. 🤔

Glease: this poor incomplete change isn't even meant to made it into the PR. it belongs to another project that attempt to extract gt++ asm to another mod. thanks for pointing out.

Glease: In all regards it feels bad to fork a oss project and not able to contribute back.. I'll still attempt to reduce the amount of breaking changes done, but will not hesitate to do so if the situation requires it.

Dream-Master (approved):
live on zeta

PR #283, opened 2022-08-18
GTpp machines pa fix

BlueWeabo:
Makes it possible for simple washer and dehydrator to work in the PA again, with the new PA changes
Requires: https://github.com/GTNewHorizons/GT5-Unofficial/pull/1272 to be merged


draknyte1 (changes requested):
Dehydrator is intentionally not supported in PAs.

BlueWeabo:
How is it not internally supported by the PA?

draknyte1:
It has a multi that’s gated appropriately.

BlueWeabo:
Then why do single block dehydrators exist?

draknyte1:
For recipes pre ZPM?
Of which dehydrators cover up to. 

Vacuum furnace is dehydrator multi, it should be the only option beyond singles. Why support has never been added, it is intentional.

BlueWeabo:
I have only made the PA possible to do the SB dehydrator recipes, from what I tested it is not possible to do the MB recipes.

draknyte1:
> I have only made the PA possible to do the SB dehydrator recipes, from what I tested it is not possible to do the MB recipes.

Because the vacuum furnace has two modes, One for dehydration. PA should not do dehydration, simple.

BlueWeabo:
If you could, can you point towards the issue, where the dehydrator was not allowed to be made put in the Processing Array? From what I know, it wasn't put there because of it's tiers are handled in its unlocalized name. Not to mention that currently it only works up to the HV recipes, pretty sure most of them are better than HV, as it only does the basic dehydrators and not the chemical+vacuumfurnace.

draknyte1:
> If you could, can you point towards the issue, where the dehydrator was not allowed to be made put in the Processing Array? From what I know, it wasn't put there because of it's tiers are handled in its unlocalized name. Not to mention that currently it only works up to the HV recipes, pretty sure most of them are better than HV, as it only does the basic dehydrators and not the chemical+vacuumfurnace.

I intentionally never added support to the PA when I added the simple washer are there are dehydrators past HV and most recipes require those. You should be using low heat mode in the Utupu Tanari if you want parallel processing. There is no ticket, it’s a design choice. 

Not everything belongs in the PA.

Glease:
To put it other way, for alk, not allowing dehydrator in PA is not a __technical limitation__, but a __game design consideration__.

BlueWeabo:
Alright. I have removed it.

draknyte1 (approved):

PR #287, opened 2022-08-20
Fix NPE with reflection

miozune:
Close https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/10429

PR #288, opened 2022-08-20
Fix bad effect removal on obfuscated environment

miozune:
See https://github.com/GTNewHorizons/GTplusplus/pull/287


Connor-Colenso (approved):
Screw obfuscation.

PR #289, opened 2022-08-22
Fix description of speed boost in tooltips

ChromicRedBrick:
The descriptions "speed: x%" in GT++ multis' tooltips have different meanings and some of them are wrong.

- For cryogenic freezer and matter fabrication CPU, x% is the speed ratio of GT++ multis to corresponding GT machines.
- For other multis, x% is how much they are faster than corresponding single block machines or GT multis.
- The tooltip of industrial sledgehammer is wrong. It actually runs 100% faster than single block forge hammers, and can do 8*AnvilTier*Tier parallel.
- Amazon warehousing depot actually runs 500% faster than single block packagers, which is not clearly mentioned in tooltip.

This pull request intends to:

1. modify the "speed: x%" in GT++ multis' tooltips to "speed: +x%", making it easier to understand
2. fix industrial sledgehammer's tooltip
3. update amazon warehousing depot's tooltip

### DIFF
--- a/src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/processing/GregtechMetaTileEntity_IndustrialChisel.java
+++ b/src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/processing/GregtechMetaTileEntity_IndustrialChisel.java
@@ -66,7 +66,7 @@ public class GregtechMetaTileEntity_IndustrialChisel extends GregtechMeta_MultiB
 				.addInfo("Factory Grade Auto Chisel")
 				.addInfo("Target block goes in GUI slot")
 				.addInfo("If no target provided, firdt chisel result is used")
-				.addInfo("Speed: +200% | Eu Usage: 75% | Parallel: Tier x 16")
+				.addInfo("Speed: +200% | EU Usage: 75% | Parallel: Tier x 16")
 				.addPollutionAmount(getPollutionPerSecond(null))
 				.addSeparator()
 				.beginStructureBlock(3, 3, 3, true)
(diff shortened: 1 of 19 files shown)


src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/processing/GregtechMetaTileEntity_IndustrialChisel.java
```diff
@@ -66,7 +66,7 @@ protected GT_Multiblock_Tooltip_Builder createTooltip() {
 				.addInfo("Factory Grade Auto Chisel")
 				.addInfo("Target block goes in GUI slot")
 				.addInfo("If no target provided, firdt chisel result is used")
-				.addInfo("Speed: 200% | Eu Usage: 75% | Parallel: Tier x 16")
+				.addInfo("Speed: +200% | Eu Usage: 75% | Parallel: Tier x 16")
```
Connor-Colenso: Can you capitalise Eu to EU while you are here? Same for the other tooltips.

Dream-Master:
@ChromicRedBrick

src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/processing/GregtechMetaTileEntity_IndustrialChisel.java
ChromicRedBrick: Done.

Also I wonder if Adv DT's parrellel is related to Tower Tier in DTower Mode. It seems the parrellel is 4 * PowerInputTier either in T1 or T2. If so, tooltip needs to be updated.

Dream-Master (approved):

PR #293, opened 2022-08-22
ABS-XL

draknyte1:
Reverts #257 & #259.
Adds the base work for the ABS-XL (@Steelux to do structure)
Made Laurenium less pink.
Fixed a stupid hard dep.
Fixed project not launching in debug through an IDE.


draknyte1:
Still requires structure code to be committed by @Steelux8

boubou19:
usually we squash the commits when we merge, not before...

draknyte1:
> usually we squash the commits when we merge, not before...

I accidentally committed everything with my work credentials, so copy pasting all changed source was easier than fixing 7 commits. It wasn’t anything useful that couldn’t have been done in a single commit anyway.

chochem:
I think when you did the reverts, you also reverted the latest structure lib/autoplace integration on the ABS that glee8e is doing (well and currently fixing).

draknyte1:
> I think when you did the reverts, you also reverted the latest structure lib/autoplace integration on the ABS that glee8e is doing (well and currently fixing).

Possibly, but that’s easy enough to put back. Oops.

boubou19:
this PR won't make it as long as the #257 and #259 are reverted. It is just adding tedium for the sake of it, it's not interesting gameplay wise, it's incoherent with the other multies and more importantly, it's **_against our vision_**.

Also, i have the approval of @Dream-Master to block this PR until conditions are met.

draknyte1:
> this PR won't make it as long as the #257 and #259 are reverted. It is just adding tedium for the sake of it, it's not interesting gameplay wise, it's incoherent with the other multies and more importantly, it's **_against our vision_**.
> 
> Also, i have the approval of @Dream-Master to block this PR until conditions are met.

Cool. I don’t see why there’s drama here. I didn’t revert anything, I rolled back to an older class version to get the old code and forgot to fix that. It’s a two second job to fix and y’all acting like it’s the apocalypse. It’s not adding tedium by any means?

257/259 are intentionally reverted to allow for this multi. The changes made after the fact by @Glease are not.

boubou19:
alk:
> Reverts https://github.com/GTNewHorizons/GTplusplus/pull/257 & https://github.com/GTNewHorizons/GTplusplus/pull/259.

also alk:
>  I didn’t revert anything, I rolled back to an older class version to get the old code and forgot to fix that.

well anyways. The condition is crystal clear: no fix for the reverts, no merge.

draknyte1:
The unintended code removed is now back.
The XL ABS exists as an upgrade path to move beyond the original multi. (As others have)
It gains Parallels & Bus splitting but at a cost, like everything else.

boubou19:
What about the code of @xSkewer ? because the whole blocking point is to allow bus separation on the normal ABS, like it was before this PR.

draknyte1:
> What about the code of @xSkewer ? because the whole blocking point is to allow bus separation on the normal ABS, like it was before.

This was never intended on the ABS and it was a design choice from the start.
You really should never need more than 7, as you only need *one* for circuits 3-9 and even then not all of them always.
A more convinient option for later game players shouldn't just be dumped onto the existing multi. It should either be through an Upgrade chip (Like Distillus) or with a better XL version which are aimed specifically at ZPM+.

boubou19:
Looks like it's not clear enough for you: **__You__** never intended to allow the bus separation. But **__we, the GTNH team__** intended it. As i said above denying bus separation is
> just adding tedium for the sake of it, it's not interesting gameplay wise, it's incoherent with the other multies and more importantly, it's against our vision.

That's it. If you want to add code support to help @Steelux8 to design the mega ABS, just make the new class and register the multi in the game. Rest is some extra bloat you are trying to force on us.

This is the GTNH fork of GT++. We are thankful for all of your hard work, you can still give us some insight about GT++, but you aren't the owner of this repository. You can't force us anymore to accept your weird choices.

repo-alt:
I'm sorry to shatter a bit that picture of uniform resolve but I have to play Olafe a bit here: I see ABS as a moderately interesting challenge to automate, and there is absolutely no need to spam 8 of them.

Steelux8:
I too agree that the super input bus automation option seems interesting to me, and the uniqueness of it was the main reason for me to have Indalloy as an ABS recipe. Unfortunately, and as was said to me by people in the past, it just is easier to build more ABS and run one circuit on each, which defeats the entire point of the automation. I didn't want to make it as trivial as all the other multis, but I preferred to just let it happen over time.

Connor-Colenso:
The issue with the super bus is it struggles to exist without causing extraordinary lag. I tried it on eta before and it can regularly take 1ms which is absurdly high for a single hatch.

Regardless of your views on this topic this reversion does not belong in this PR and should be split into another one. But given they were recently merged I don't see why we need to go over this again tbh.

repo-alt:
You don't really need superbus though. The bus separation (and casings reduction) main justification (as I see it) is that any ABS automation is definitely not one tick that you usually require at late game (because of ridiculous epeen pack design (as Lea would not say)), but since mega version would definitely have many inputs that issue is no more.
To automate regular ABS you can you a microcontroller (I used a transposer but it is much more cumbersome), or you can (probably) use SFM. Current microcontroller code I use (borrowed from FalseColor) is 50 lines of code (comments and blank lines included)

Steelux8:
I'm just about done with my part of implementing this multi. The remaining issues are as such:

- [ ] Conflicting files;
- [ ] Naquadah Containment Casing has no recipe, I chose it as the low tier version of the Matter Fabrication Casing;
- [ ] TecTech multi-amp energy hatches do not work. I checked the MBF code, and it seems to me that GT++ needs to depend on TecTech for that. I decided to leave this to a more competent dev;
- [ ] The bonuses when upgrading the casings are still undecided. I'm nowhere near ZPM+, the tier of this multi, so I have no way of properly balancing the bonuses with the costs;
- [ ] The recipe map seems to be built properly, and there are methods to add to this multi exclusively, in the same way that recipes would be added to the ABS originally. However, NEI does not recognize the recipe page for this multi properly;
- [ ] Other balancing decisions, such as the amount of parallels, speed boosts and overclocking/perfect overclocking still need to be made.

boubou19:
superseded by https://github.com/GTNewHorizons/GTplusplus/pull/394

PR #294, opened 2022-08-22
fix a bunch of structure code error

Glease:
probably not the last of this series...

### DIFF
--- a/src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/processing/GregtechMetaTileEntity_IndustrialArcFurnace.java
+++ b/src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/processing/GregtechMetaTileEntity_IndustrialArcFurnace.java
@@ -106,7 +106,7 @@ public class GregtechMetaTileEntity_IndustrialArcFurnace extends GregtechMeta_Mu
 					.addElement(
 							'C',
 							buildHatchAdder(GregtechMetaTileEntity_IndustrialArcFurnace.class)
-									.atLeast(InputBus, OutputBus, Maintenance, Energy, Muffler)
+									.atLeast(InputBus, InputHatch, OutputBus, Maintenance, Energy, Muffler)
 									.casingIndex(getCasingTextureIndex())
 									.dot(1)
 									.buildAndChain(onElementPass(x -> ++x.mCasing, ofBlock(ModBlocks.blockCasings4Misc, 3)))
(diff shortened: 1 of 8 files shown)


src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntity_AlloyBlastSmelter.java
```diff
@@ -117,7 +117,7 @@ public IStructureDefinition<GregtechMetaTileEntity_AlloyBlastSmelter> getStructu
 					.addElement(
 							'C',
 							buildHatchAdder(GregtechMetaTileEntity_AlloyBlastSmelter.class)
-									.atLeast(InputBus, OutputBus, Maintenance, Energy, Muffler)
+									.atLeast(InputBus, InputHatch, OutputBus, Maintenance, Energy, Muffler)
```
chochem: needs Output hatch too

Glease: Actually, isn't this fluid output only?

chochem: needs both, for example for rhenium

src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/processing/GregtechMetaTileEntity_IndustrialArcFurnace.java
```diff
@@ -106,7 +106,7 @@ public IStructureDefinition<GregtechMetaTileEntity_IndustrialArcFurnace> getStru
 					.addElement(
 							'C',
 							buildHatchAdder(GregtechMetaTileEntity_IndustrialArcFurnace.class)
-									.atLeast(InputBus, OutputBus, Maintenance, Energy, Muffler)
+									.atLeast(InputBus, InputHatch, OutputBus, Maintenance, Energy, Muffler)
```
chochem: I think this also needs an output hatch when used as plasma arc furnace

chochem (approved):
have not looked at all of them, but volcanus, cutting machine, arc furnace, and ABS are fixed

PR #300, opened 2022-08-25
Fix Pipes flowing rate calculation

### LINKED ISSUE #11103: 4x, 9x and 16x bronze fluid pipes have inconsistent throughput

NotAPenguin#5179

### Your Pack Version

dev18

### Your Server

SP

### Type of Server

Single Player

### Your Expectation

Each pipe has the same throughput

### The Reality

image

The nonuple pipe has a different throughput value than the others.

### Your Proposal

Make the pipe throughput consistent across all variants

### Final Checklist

Quarri6343:
Close https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/11103 by adding missing per second to per tick division to hexadecuple pipes flowing rate.
Also, 1 second is supposed to be 20 ticks so I also fixed that weird division.

### DIFF
--- a/src/main/java/gtPlusPlus/xmod/gregtech/registration/gregtech/GregtechConduits.java
+++ b/src/main/java/gtPlusPlus/xmod/gregtech/registration/gregtech/GregtechConduits.java
@@ -114,11 +114,12 @@ public class GregtechConduits {
 
 	}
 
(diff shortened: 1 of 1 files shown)


miozune (approved):
they might be a bit too weak, but given we don't even have recipes for them it's fine

PR #301, opened 2022-08-25
Fix void protection requires too many slots than actual

miozune:
You can check with Froth processing with Vacuum Furnace


src/main/java/gtPlusPlus/xmod/gregtech/api/metatileentity/implementations/base/GregtechMeta_MultiBlockBase.java
```diff
@@ -489,24 +486,17 @@ public int canBufferOutputs(ItemStack[] aItemOutputs, FluidStack[] aFluidOutputs
...
-					else {
-						// Only requires one slot
-						aRecipeSlotsRequired++;
+
+					int aSlotsNeedsForThisStack = (int) Math.ceil((float) aStackSize / aY.getMaxStackSize());
+					// Should round up and add as many stacks as required nicely.
+					for (int o=0;o<aSlotsNeedsForThisStack;o++) {
+						if (aStackSize < 1) break;
+						int aStackToRemove = Math.min(aStackSize, aY.getMaxStackSize());
+						aStackSize -= aY.getMaxStackSize();
```
minecraft7771: Shouldn't cause issues but shouldn't you decrease `aStackSize` it by `aStackToRemove`  instead of the max stack size?

miozune: Oh yes true

minecraft7771 (approved):

PR #258, opened 2022-07-31
Fix startup crash with latest GT

miozune:
New constructors with `GT_MetaTileEntity_Hatch_Energy` and `GT_MetaTileEntity_Hatch_Dynamo` don't work with ASMed ones. Removes them and uses reflection instead.

PR #303, opened 2022-08-25
Fix reflection failure with RTG hatch

miozune:
See https://github.com/GTNewHorizons/GTplusplus/pull/258
I'm not sure why `ReflectionUtils` fails so I just wrote plain Reflection


eigenraven (changes requested):
Can you just make a PR to GT to make the field not final, and then you could use simple assignment here?

miozune:
I think reassigning `mInventory` is super rare situation and removing `final` should be avoided as much as possible

eigenraven (approved):

PR #309, opened 2022-08-26
Fix assembler recipe for Wither Cage being more expensive than crafti…

### LINKED ISSUE #7846: GT++ Wither cage assembler recipe more expensive than crafting table recipe
2.1.0.6
#
Private Server
#
#### Describe the Issue
The Crafting table requires fewer materials than the assembler recipe with the same output.
#
#### Screenshots
image
image
#

Change the assembler recipe to either require the same materials for the same output or make the assembler recipe more efficient rather than less.
#
#### What is your GTNH Discord username?
`Dr. Turtle#0001`

miozune:
…ng recipe
Close https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/7846


minecraft7771 (commented):
The assembler recipe now requires half as much nether stars. Is that an intendend change?

miozune:
I guess that's what Alk wanted to but just a guess

Dream-Master (approved):

PR #317, opened 2022-08-29
a less buggy and more performant canBufferOutput implementation for fluid management

Glease:


src/main/java/gtPlusPlus/xmod/gregtech/api/metatileentity/implementations/base/GregtechMeta_MultiBlockBase.java
```diff
@@ -505,8 +506,7 @@ public int canBufferOutputs(ItemStack[] aItemOutputs, FluidStack[] aFluidOutputs
 
             // A map to hold the items we will be 'inputting' into the output buses. These itemstacks are actually the
             // recipe outputs.
-            ConcurrentSet<FlexiblePair<ItemStack, Integer>> aInputMap =
-                    new ConcurrentHashSet<FlexiblePair<ItemStack, Integer>>();
+            Set<FlexiblePair<ItemStack, Integer>> aInputMap = new HashSet<>();
```
kuba6000: Why is it change to normal HashSet ? Later in the code it is used in normal loop and elements are removed
(https://github.com/GTNewHorizons/GTplusplus/pull/317/files#diff-155b83f6cf02853ae3ea8cf128c71abf6ed7fb3b2f7d7b43d62402209e8179c6R557)
The loop should be change to iterator loop or that should be reverted.

kuba6000 (changes requested):

src/main/java/gtPlusPlus/xmod/gregtech/api/metatileentity/implementations/base/GregtechMeta_MultiBlockBase.java
Glease: Good catch. Thanks.

src/main/java/gtPlusPlus/xmod/gregtech/api/metatileentity/implementations/base/GregtechMeta_MultiBlockBase.java
```diff
@@ -566,7 +568,7 @@ public int canBufferOutputs(ItemStack[] aItemOutputs, FluidStack[] aFluidOutputs
                                     FlexiblePair<ItemStack, Integer> t =
                                             new FlexiblePair<ItemStack, Integer>(u.getKey(), aRemainder);
                                     // Remove the 'input' stack from the recipe outputs, so we don't try count it again.
-                                    aInputMap.remove(u);
+                                    iterator.remove();
                                     // Add the remainder stack.
                                     aInputMap.add(t);
```
kuba6000: I'm pretty sure that addition will throw CME.
Maybe that continue should continue bus iteration

src/main/java/gtPlusPlus/xmod/gregtech/api/metatileentity/implementations/base/GregtechMeta_MultiBlockBase.java
```diff
@@ -554,7 +556,7 @@ public int canBufferOutputs(ItemStack[] aItemOutputs, FluidStack[] aFluidOutputs
                                     // Update the stack size in the bus storage map.
                                     y.setValue(aOutputBusStack.stackSize + aOutputStack.stackSize);
                                     // Remove the 'input' stack from the recipe outputs, so we don't try count it again.
-                                    aInputMap.remove(u);
+                                    iterator.remove();
                                     continue outputItems;
```
kuba6000: There could be a check to continue bus iteration when the stack is full
Also a few lines before (L550) can be moved before stack equal check for some optimization as well.

kuba6000 (changes requested):

src/main/java/gtPlusPlus/xmod/gregtech/api/metatileentity/implementations/base/GregtechMeta_MultiBlockBase.java
```diff
@@ -473,306 +477,203 @@ public int canBufferOutputs(ItemStack[] aItemOutputs, FluidStack[] aFluidOutputs
...
-                if ((aFreeHatchCheck.getValue_2() == null
-                                || aFreeHatchCheck.getValue_3() == 0
-                                || aFreeHatchCheck.getValue_1().getFluid() == null)
-                        && !aFreeHatchCheck.getValue_1().isFluidLocked()) {
-                    aEmptyFluidHatches++;
-                }
-            }
+        // check non-restrictive hatches
+        for (GT_MetaTileEntity_Hatch_Output tHatch : mOutputHatches) {
+            int tSpaceLeft = tHatch.getCapacity() - tHatch.getFluidAmount();
```
kuba6000, line 638: Making second loop can be avoided by sorting mOutpuHatches. Its not really that big of a problem I guess

kuba6000 (approved):
Code looks ok.

src/main/java/gtPlusPlus/xmod/gregtech/api/metatileentity/implementations/base/GregtechMeta_MultiBlockBase.java
Glease: I thought about sorting. But the fact I have to sort these every time this is called dissuade me from doing so, and I don't want to deviate from the original addOutputs() method in base class too much, as otherwise we run the risk of giving false negatives in some unforseen ways.

kuba6000: Huh now when I think about it, the loop can be removed anyway. Order doesn't matter, the function only checks if you can output them. The output really happens in GT5 code and its outputting to the right hatches.

Glease: Ideally I prefer to not have this function to return true if addOutputs would void.

--

The case in my head is when player are mixing hatches with different tier. Order can be significant in that case. I'm honestly not sure this case is valid, but I prefer to stay on the safe side until I'm willing to actually grab a pencil and paper to prove it.

PR #339, opened 2022-09-08
radhatch compact

GlodBlock:
make rad hatch accept a2's radioactive material

### DIFF
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -1,5 +1,5 @@
 dependencies {
-    compile('com.github.GTNewHorizons:GT5-Unofficial:5.09.41.27:dev')
+    compile('com.github.GTNewHorizons:GT5-Unofficial:5.09.41.33-pre:dev')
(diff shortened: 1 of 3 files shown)


draknyte1 (commented):
Cyclic dependency with BW isn’t favourable. Plus the support should be added there anyway?

GlodBlock:
it is only a load check so there is no cyclic dependency

draknyte1:
> it is only a load check so there is no cyclic dependency

It depends on BW classes and from memory BW already depends on GT++. Why not just add this to BW?

Dream-Master:
@GlodBlock @draknyte1 is this possible and why it would be better ?

draknyte1:
RadHatch is part of BW and if it wants to support content from other mods it already has as a dependency, it makes the most sense to put it there.

GlodBlock:
will be moved to coremod

PR #345, opened 2022-09-13
Bugfixes and Speed Increases for the LFTR Chain

Steelux8:
- Changed a few restrictions to allow this chain to be doable in IV, as intended;
- Changed the EU/t of the 3 fuels to match the usage of 4 buffered dynamos, and reduced the burning time to keep the total energy per L the same;
- Lowered the recipe time of some recipes done in multiblocks;
- Other tooltip and bug fixes.

Some bugs that this chain had should now be fixed. Additionally, the EU/t of the LFTR was significantly increased, as it was designed before buffered hatches were a thing, and as such one of these expensive 7x7 multis would output exactly as much as a single turbine multi with a buffered IV hatch. With this change, Fuel 1 equates to 4 buffered EVs, Fuel 2 to 4 buffered IVs and Fuel 3 to 4 buffered LuVs. This also adds another factor of progression to the chain, as upgrading 4 dynamos is expensive when early in the tier.

The other balance change was in the noble gas output from sparging with Helium, which has increased significantly. Now, Helium can be converted to 5 other, better noble gases (up to Xenon) at a rate of around 100L per 25s for every gas. This is a lot slower than gas siphons on other planets, but it's available earlier and consumes a lot less EU, especially compared to the Xenon siphoning.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,4 +1,4 @@
-//version: 1661114848
+//version: 1662920829
 /*
  DO NOT CHANGE THIS FILE!
  Also, you may replace this file at any time if there is an update available.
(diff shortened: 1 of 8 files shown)


Steelux8:
image

This message does pop up after the change, because I removed the dust of Ammonium Tetrafluoroberyllate, which caused a recipe conflict and also isn't used for anything. If anyone wishes to catch this missing exception to prevent the error from showing up, I'd be grateful.

Dream-Master (approved):

Anarack:
The sparging recipes in game are not correct, PR description indicates they should be 25s for 100L of sparged gas, instead in game they are 250 seconds for 200 L of the associated noble gas in nei. 
image

Edit NM, maybe tool tip or something should be adjusted to show that the tower is 10x fast?

PR #347, opened 2022-09-13
Rocketdyne Structure Fix and Cost Reduction

Steelux8:
- Fixed Air Intake Hatches not being detected by the controller;
- Reduced the amount of Tungstensteel needed by replacing some with Titanium.

The Rocketdyne was unable to form, because of the survival construct PR which removed a method essential for it to count the amount of Air Intake Hatches. I brought that method back.

Additionally, the multi needs a minimum of 64 Turbodyne Casings, but each one costs 20 Tungstensteel Ingots, so building this would cost 20 stacks of TSS just for these casings, adding to the rest of the items needed. To make it cheaper, I replaced the pipes with Titanium ones and added a Long Titanium Rod in the previously empty slot, which means it now costs 6 TSS stacks and 13 Titanium stacks, plus the other materials, to make these casings.

### DIFF
--- a/src/main/java/gtPlusPlus/xmod/gregtech/registration/gregtech/GregtechRocketFuelGenerator.java
+++ b/src/main/java/gtPlusPlus/xmod/gregtech/registration/gregtech/GregtechRocketFuelGenerator.java
@@ -57,11 +57,13 @@ public class GregtechRocketFuelGenerator {
             "RFR",
             "PWP",
             'R',
-            OrePrefixes.pipeLarge.get(Materials.TungstenSteel),
+            OrePrefixes.pipeLarge.get(Materials.Titanium),
             'F',
             ItemList.Casing_RobustTungstenSteel,
             'P',
-            ALLOY.NITINOL_60.getGear(1)
+            ALLOY.NITINOL_60.getGear(1),
+            'W',
+            OrePrefixes.stickLong.get(Materials.Titanium)
         });
 
         GregtechItemList.Rocket_Engine_EV.set(new GregtechMetaTileEntityRocketFuelGenerator(
--- a/src/main/java/gtPlusPlus/xmod/gregtech/registration/gregtech/GregtechRocketFuelGenerator.java
+++ b/src/main/java/gtPlusPlus/xmod/gregtech/registration/gregtech/GregtechRocketFuelGenerator.java
@@ -57,13 +57,13 @@ public class GregtechRocketFuelGenerator {
             "RFR",
             "PWP",
             'R',
-            OrePrefixes.pipeLarge.get(Materials.Titanium),
+            OrePrefixes.pipeMedium.get(Materials.TungstenSteel),
             'F',
             ItemList.Casing_RobustTungstenSteel,
             'P',
             ALLOY.NITINOL_60.getGear(1),
             'W',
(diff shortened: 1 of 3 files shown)


Glease (commented):
1. I'll do the structure fix myself. 
2. I don't see why on earth do you want this LuV+/ZPM+ generator to use titanium. The `W` fix is good though.

Dream-Master:
> * I'll do the structure fix myself.
> * I don't see why on earth do you want this LuV+/ZPM+ generator to use titanium. The `W` fix is good though.

I would not change it the pipe to Titanium. Gt++ has so many tiers mixed in a recipe like material tier x circuits tier y casing tier z.

boubou19:
tungstensteel/titanium are like dirt in LuV/ZPM so i don't really understand the change of the cost

Steelux8:
> 2. I don't see why on earth do you want this LuV+/ZPM+ generator to use titanium. The `W` fix is good though.

What makes this multi strictly LuV+? I didn't see anything that places it above IV.

Dream-Master:
Even in IV you can make Tungstensteel.

King65868:
You can make tungstensteel at IV, but no one one is going to spend 20 stacks of it in IV on the multi

Dream-Master:
Then build it later. Since when we balance the pack with arguments like this above?

Glease:
Because its output range is in the tier of LuV-ZPM? I do not yet know anyone plan to use these to power their IV base yet.

src/main/java/gtPlusPlus/xmod/gregtech/registration/gregtech/GregtechRocketFuelGenerator.java
```diff
@@ -57,11 +57,13 @@ private static void run1() {
             "RFR",
             "PWP",
             'R',
-            OrePrefixes.pipeLarge.get(Materials.TungstenSteel),
+            OrePrefixes.pipeLarge.get(Materials.Titanium),
```
Dream-Master: re add Tungstensteel

src/main/java/gtPlusPlus/xmod/gregtech/registration/gregtech/GregtechRocketFuelGenerator.java
```diff
@@ -57,11 +57,13 @@ private static void run1() {
...
             'F',
             ItemList.Casing_RobustTungstenSteel,
             'P',
-            ALLOY.NITINOL_60.getGear(1)
+            ALLOY.NITINOL_60.getGear(1),
+            'W',
+            OrePrefixes.stickLong.get(Materials.Titanium)
```
Dream-Master: and if you add this part change it to Tungstensteel as well

Dream-Master (changes requested):

Steelux8:
I changed the pipes and the long stick to TSS, but I also changed the pipes to medium instead of large, so that the total amount of TSS stacks for the casings is 15 instead of 21.

The math is favorable towards this multi being expensive, and GT++ rocket fuels being the only craftable ones should give it a sufficient boost to see proper usage and replace LCE/LGT spam.

Dream-Master (approved):

PR #373, opened 2022-09-25
Revert "Fix XL plasma turbine consuming way too much plasma"

minecraft7771:
Reverts GTNewHorizons/GTplusplus#372


chochem (approved):
not tested but should be the same as the base turbines

PR #387, opened 2022-10-03
Remove duplicate keys

ChromicRedBrick:
Duplicate keys make no sense.


eigenraven (dismissed):
Remove instead of commenting

Dream-Master (approved):

PR #391, opened 2022-10-04
handle drops with a smaller max stack size properly

Glease:
mostly to address primordial pearl issues.

also inverted a few if statements to reduce nesting levels. the effect and semantic is left the same


Glease:
example

boubou19:
i like how the gt++ code is shrinking more and more while becoming cleaner after each PR 😂

minecraft7771 (approved):

PR #394, opened 2022-10-06
Mega Alloy Blast Smelter

MadMan310:
Aimed at UV-Tier.
Up to 256 parallels.
Same parallel logic as the other mega multiblocks.

The coils give a speed bonus, based on the function TIER/130:
image
The reason why is because, in practice, there are really 14 tiers of coils, starting from 0 (due to the "existance" of the None and ULV tier coils)

image
**(Ignore the 5 second recipe time, its actually 60 seconds)**

Because @S4mpsa did not like efficiency bonuses, normal ABS coils can be used in their place, giving no effect on speed.

The borosilicate glass limits the energy hatch tier. UV+ allows TT laser hatches. UEV+ unlocks all other tiers.

Supports:
-Overclocking beyond MAX
-Separate buses, use shift+screwdriver to toggle

This was created by me due to various disagreements with Steelux's "ABS-XL" from me, Phineasor , Colen, S4mpsa, GDCloud, and other endgame players.


chochem (dismissed):
Such a PR needs a lot more information in the description. It also needs to be discussed first. I am against it, as we already have a (imo much more interesting) XL/mega ABS in the works. see https://github.com/GTNewHorizons/GTplusplus/pull/293.

boubou19:
in addition to that, i don't like the concept of dragging BW in GT++ deps as iirc BW makes GT++ as a dep. let's avoid circular deps

Steelux8:
More BW-styled megas are far from a good addition. I put in a lot of work to customize the structure and implementation of the mega ABS I was working on, and I'd much prefer if that one was added and, in the future, the current megas changed to be more creative and customizable.

chochem:
Ok maybe I am not that strongly against it, looking how many problems the other PR faces. But please provide a lot more info. (picture, mechanics, tiering or recipe, etc.)

One specific question from a quick glance at the code: Whats up with the very strange discount formula? actually thats more like 4 question. 1) the tooltip and the code disagree on what the formula is 2) the tooltip formula is just plain undefined for tier 1/2/3  3) why use this weird constant log2_40 instead of just lets say 30. and most importantly 4) whats the motivation of the logarithmic scaling, why not just make it linear? linear would be a lot easier to communicate and more meaningful. Can also go a bit higher imo but thats a separate (balancing) question.

Phineasor:
> 

@chochem The reason the the strange discount formula, is that @S4mpsa did not like it when it was a linear discount, because it would make a 20 second recipe get done in 9 seconds, with maximum coils and that was too much, so madman changed it so that max speed is 10 percent faster @MadMan310  please let me know if im wrong. And ya, the tool tip is wrong, or the check is wrong, ether way that needs to be fixed.

@boubou19  I agree, cric dep bad, can it be removed by moving this to BW?

EDIT: I just checked, and with my limited, but currently expanding knowledge of Java I think Madman already fixed the discount disagreement.

chochem:
> The reason the the strange discount formula, is that @S4mpsa did not like it when it was a linear discount, because it would make a 20 second recipe get done in 9 seconds, with maximum coils and that was too much, so madman changed it so that max speed is 10 percent faster

Those two things have nothing to do with each other. You can just have a reasonable maximum value in a linear formula.

It sounds like I missed some prior discussions, but I mean thats why you put info into a PR for everyone.

src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/mega/GregTechMetaTileEntity_MegaAlloyBlastSmelter.java
```diff
@@ -0,0 +1,620 @@
...
+    }
+    
+    @Override
+    protected GT_Multiblock_Tooltip_Builder createTooltip() {
+        GT_Multiblock_Tooltip_Builder tt = new GT_Multiblock_Tooltip_Builder();
+        tt.addMachineType("Fluid Alloy Cooker")
+        .addInfo("Controller block for the Mega Alloy Blast Smelter")
+        .addInfo("Runs the same recipes as the normal ABS, except with up to "+EnumChatFormatting.BOLD+EnumChatFormatting.UNDERLINE+MAX_PARALLELS+EnumChatFormatting.RESET+EnumChatFormatting.GRAY+" parallels.")
+                .addInfo("Every coil tier above cupronickel grants a speed bonus, based on this function:")
+                .addInfo("log(TIER - 3) / log(2^40), rounded to the nearest 10-thousandth. This caps out at 10%.")
```
chochem: This is the incorrect formula to be more specific. probably means Tier+3.

Phineasor: @chochem I see what you mean, it would be easier to just make it so that a linear function ends at the correct maximum value that would be the better solution so that the earlier coil tiers are not a larger upgrade than the later ones. @MadMan310  please fix this

@chochem  I see what you mean with that disagreement now, thank you for pointing out exactly where  I did not notice that the +- was wrong, however the resolution of this formula will require a rework of this anyways.

boubou19: btw, log(2^40) = 40*log(2), which is less confusing for people uncomfortable with maths.

Connor-Colenso: Please be clear on which base log as well, it appears Math.log is log base e, not base 2 or 10. In which case, use ln not log in descriptions.

MadMan310: @boubou19 I don't see any reason to remove the BW dependency. You are concerned about a circular dependency, yet BW and GT++ had no dependencies on each other in the first place: https://github.com/GTNewHorizons/bartworks/blob/ff8a1e590bea7b78215a21c8f90c9955aa2ebb4c/dependencies.gradle

MadMan310: @Steelux8 I have a preposition: Your multiblock could be used for any new and much more advanced alloys; the ABS-XL would need to be renamed if that were to be the case however. The Mega ABS would then be used for all currently existing Alloy Blast Smelter recipes.

Connor-Colenso:
I still think the mufflers should be raised up a block, makes the top less flat and they will take the texture of the casing anyway. But just my opinion 😛.

Steelux8:
Why is the recipe so expensive and so lategame? If this is meant to copy the other megas, it should also copy how early they are made available.

Phineasor:
@Steelux8  its a mid UV multi, because that is when you start to need to use this much ABS power, it also has a more interesting and reasonable recipe than 64 controllers. and when you really start to need this(at least currently, hopefully more uses for large scale GTPP alloys will get added between UV and UEV) is the DTPF just after UEV.

Steelux8:
But this *is* a BW-styled mega. I already have worked plenty on a bigger ABS that doesn't follow the BW mega's usual design choices, so this one should mimic the recipes and earliest tier of the MBF, mega vac freezer, mega DT, etc.

Connor-Colenso:
If anything, the other megas need better recipes. 64x controllers + soldering alloy is just lazy.

Phineasor:
@Steelux8 I agree with connor here we should just fix the other mega controllers so they are less boring, also I have had some ides for uses for your structure, for possibly later game features if needed, because to my knowledge your machine is more than just an ABS like this one is. in my mind this is more of an industrial production machine where the regular ABS is a machine used for prototyping alloys, and perhaps yours can do something more advanced.

Steelux8:
> If anything, the other megas need better recipes. 64x controllers + soldering alloy is just lazy.

I agree with this, but if this is the way to go, then this mega ABS should not have the coils, because that has nothing to do with the ABS and is taken directly from the MBF. The tiered glass is also unrelated to the ABS, but the EBF doesn't have it either so that's not too bad.

And yes, I would like the multi I designed to be more than just a mega multi. Unfortunately, I don't have experience in balancing for those tiers.

Phineasor:
@Steelux8  in my mind because the multi is bigger there is space for those coils that help the MABS heat up faster(thus the slight benefit they provide to speed but you can use the regular abs coils if you feel), because there then are the regular coils in the structure as well to do their purpose (in my mind while these coils produce heat they also heat in such a way that convection occurs to mix the fluids together and make the alloys). (plus it makes the mega more expensive so that its not absurdly cheap)

Edit, I read your post on the XL abs pr you don't have experience at that tier yet ether(I hope you get there soon! zpm is fun) so it seems other people would be needed to balance that stuff anyways(if it does get used for post DTPF content I guess its me and GD, plus it would look better at that tier)

Connor-Colenso (approved):
Seems reasonable to me and it has been play tested in full pack by Phin.

PR #396, opened 2022-10-09
Make Chemplant Mv and use mv materials

Dream-Master:
Make Chemplant Mv and use mv materials

### DIFF
--- a/src/main/java/gtPlusPlus/core/recipe/RECIPES_Machines.java
+++ b/src/main/java/gtPlusPlus/core/recipe/RECIPES_Machines.java
@@ -1026,13 +1026,13 @@ public class RECIPES_Machines {
         CORE.RA.addSixSlotAssemblingRecipe(
                 new ItemStack[] {
                     CI.getNumberedBioCircuit(19),
-                    CI.getTieredGTPPMachineCasing(1, 6),
-                    CI.getTieredComponentOfMaterial(Materials.Steel, OrePrefixes.gearGt, 6),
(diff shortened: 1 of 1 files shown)


draknyte1 (commented):
It’s meant to be made before MV, if you’re a masochist. Why tier it later when it has recipes for lower tiers?

Dream-Master:
no one need chemplant before mv. this multi is huge i didnt not know anyne have this device in LV.

chochem:
> It’s meant to be made before MV, if you’re a masochist. Why tier it later when it has recipes for lower tiers?

didnt it already need MV parts and power? this more make the recipe more consistent and not half LV/half MV

Steelux8 (changes requested):
This PR needs a purpose. Why is this change made? Is there a need for this change?

boubou19:
the purpose of this is simple: adding consistancy to some weird crafts. if there is no use before mv then it should be mv. There is literally no reason to craft it before so it shouldn't be a significant change anyways.

boubou19 (approved):

draknyte1:
> the purpose of this is simple: adding consistancy to some weird crafts. if there is no use before mv then it should be mv. There is literally no reason to craft it before so it shouldn't be a significant change anyways.

Except it has LV recipes? 🤔
Cause ya know, Algae farm.

POPlol333:
Does Algae have even a use in LV ? I literally used them first time in UV-ish

Steelux8:
I didn't dislike the recipes. The Algae Farm can be effectively MV, I suppose.

PR #399, opened 2022-10-12
Add the Quantum Force Transformer (Endgame Chemline Skipper)

Steelux8:
**I will need help with the code and balance to get this multi ready for release.**

This is the same structure, and the same code I was working on for the mega ABS. However, someone else coded in a simpler and very effective implementation for that, so my work was kicked off the role. Instead of that, I propose that this multi is made to produce output of popular chemical lines (epoxy, PBI, platline, etc.), but in one step and without needing early-tier inputs. 

This will be gated at around UEV, with Bioware Mainframes and possibly other items such as an Eternal Singularity. Its recipes will consist of expensive, lategame inputs that will turn into the useful outputs of those chemical lines, in large batches, all to minimize lag, machine spam and reliance on early tier recipes and lines for endgame progress. Currently, however, this still has the ABS-oriented code that Alkalus wrote for the purpose of making it a mega ABS.

I implemented two tiers in three of the casings present in the multi. I'd like to see this used to make overall tiers of the multi, unlocking more and more recipes and possibly also giving some bonuses here and there. However, this was originally balanced for ZPM+, so it'll be difficult to properly tier it giving that I reused several existing casings instead of making them all brand new.

Here is a list of what needs to be done for the QFT:
- [x] Remove the code related to the ABS;
- [x] Change the recipemap code and allow for a large amount of inputs and outputs;
- [x] Lock recipes behind tiers as decided by balance, and possibly add more tiers;
- [x] Decide on recipes for the controller and casings that don't have recipes yet;
- [x] Decide what happens with the reused casings that aren't balanced for endgame;
- [x] General balance decisions towards which lines can be skipped, and what the new inputs/outputs are;
- [x] Disable animations for the multi unless it is actively running;
- [x] Add a visual, intangible force field on the open spaces around the base of the multi;
- [x] Improve some textures, particularly the dark purple casings and the "glass".

Current look of the multi:
 
GIF 01-09-2022 20-13-42


Phineasor:
This looks like it will be incredibly useful for lag reduction and adding more interesting machines, will it be a more efficient version, or just faster?

Steelux8:
> This looks like it will be incredibly useful for lag reduction and adding more interesting machines, will it be a more efficient version, or just faster?

The plan is to have it both craft in larger batches at a time, to process less recipes, and also to avoid low-tier inputs that force endgame players to carry a lot of unnecessary old setups with them.

BlueWeabo:
i am sorry i forgot ;-;

src/main/java/gtPlusPlus/xmod/gregtech/recipes/GregtechRecipeAdder.java
```diff
@@ -599,6 +599,59 @@ public boolean addBlastSmelterRecipe(
...
+        for (int i = 0; i < aFluidOutput.length; i++) {
+            if ((aDuration = GregTech_API.sRecipeFile.get(
+                            "quantumsmelter", aFluidOutput[i].getFluid().getName(), aDuration))
+                    <= 0) {
+                Logger.WARNING("Recipe did not register.");
+                return false;
+            }
+        }
+
+        for (int das = 0; das < aInput.length; das++) {
```
minecraft7771: Whats `das`? I'd suggest either using the standard `i` or a more explanatory name.

src/main/java/gtPlusPlus/xmod/gregtech/registration/gregtech/GregtechQuantumForceTransformer.java
```diff
@@ -0,0 +1,25 @@
...
+    public static void run() {
+        if (gtPlusPlus.core.lib.LoadedMods.Gregtech) {
+            Logger.INFO("Gregtech5u Content | Registering Quantum Force Transformer Multiblock.");
+            if (CORE.ConfigSwitches.enableMultiblock_QuantumForceTransformer) {
+                run1();
+            }
+        }
+    }
+
+    private static void run1() {
```
minecraft7771: This function only contains one line of code and is only used in one place, so inline it.

src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntity_QuantumForceTransformer.java
```diff
@@ -0,0 +1,1160 @@
...
+        implements ISurvivalConstructable {
+
+    private int mCasing;
+    protected int mChamberTier = 0;
+    protected int mFabCoilTier = 0;
+    protected int mGlassTier = 0;
+    protected int mMinimumMufflerTier = 0;
+    private boolean mSeparateInputBusses = false;
+    private boolean mFluidMode = false, doFermium = false;
+    private Fluid mNeptunium = ELEMENT.getInstance().NEPTUNIUM.getPlasma();
```
minecraft7771: Why not make this and `mFermium` static? It's never changed.

src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntity_QuantumForceTransformer.java
```diff
@@ -0,0 +1,1160 @@
...
+                .addStructureHint("Catalyst Housing", 4)
+                .addMufflerHatch("Top Layer (except edges), x21", 5)
+                .toolTipFinisher(CORE.GT_Tooltip_Builder);
+        return tt;
+    }
+
+    @Override
+    public IStructureDefinition<GregtechMetaTileEntity_QuantumForceTransformer> getStructureDefinition() {
+        if (this.STRUCTURE_DEFINITION == null) {
+            this.STRUCTURE_DEFINITION = StructureDefinition.<GregtechMetaTileEntity_QuantumForceTransformer>builder()
```
minecraft7771: Why is the structure definition defined here instead of a direct asignment to `STRUCTURE_DEFINITION `?

src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntity_QuantumForceTransformer.java
```diff
@@ -0,0 +1,1160 @@
...
+            IMetaTileEntity aMetaTileEntity = aTileEntity.getMetaTileEntity();
+            if (aMetaTileEntity instanceof GT_MetaTileEntity_Hatch_Catalysts) {
+                return addToMachineList(aTileEntity, aBaseCasingIndex);
+            }
+        }
+        return false;
+    }
+
+    @Override
+    public boolean addToMachineList(IGregTechTileEntity aTileEntity, int aBaseCasingIndex) {
```
minecraft7771: Maybe it would be a good idea to merge this function with `addCatalystHousing`, since they are almost identical.

src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntity_QuantumForceTransformer.java
```diff
@@ -0,0 +1,1160 @@
...
+    }
+
+    @Override
+    public boolean checkMachine(IGregTechTileEntity aBaseMetaTileEntity, ItemStack aStack) {
+        this.mCasing = 0;
+        this.mMinimumMufflerTier = 0;
+        mCatalystBuses.clear();
+        if (checkPiece(this.mName, 7, 20, 4)
+                && checkHatch()
+                && (mTecTechEnergyHatches.size() >= 1 || mEnergyHatches.size() >= 1)) {
```
minecraft7771: A nitpick but if you already want to check if enough hatches are present, why not also check for Maintenance hatches.

src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntity_QuantumForceTransformer.java
```diff
@@ -0,0 +1,1160 @@
...
+        return false;
+    }
+
+    @Override
+    public void construct(ItemStack stackSize, boolean hintsOnly) {
+        buildPiece(this.mName, stackSize, hintsOnly, 7, 20, 4);
+    }
+
+    @Override
+    public int survivalConstruct(ItemStack stackSize, int elementBudget, IItemSource source, EntityPlayerMP actor) {
```
minecraft7771: It would be good to migrate to the new survival placement calls, that use the `AutoPlaceEnvironment `.

src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntity_QuantumForceTransformer.java
```diff
@@ -0,0 +1,1160 @@
...
+        FluidStack[] tFluidList = getCompactedFluids();
+        ItemStack[] tCatalysts = mCatalystBuses.get(0).mInventory;
+        if (mSeparateInputBusses) {
+            ArrayList<ItemStack> tInputList = new ArrayList<ItemStack>();
+            for (GT_MetaTileEntity_Hatch_InputBus tBus : mInputBusses) {
+                for (int i = tBus.getSizeInventory() - 1; i >= 0; i--) {
+                    if (tBus.getStackInSlot(i) != null) {
+                        tInputList.add(tBus.getStackInSlot(i));
+                        if (ItemUtils.isCatalyst(tBus.getStackInSlot(i))) {
+                            if (mCurrentParallel < mMaxParallel) {
```
minecraft7771: Why is parallel logic here when its also in `processRecipe`, which is also called within this if-condition?

minecraft7771 (changes requested):

src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntity_QuantumForceTransformer.java
BlueWeabo: Because there are methods, which are used, but aren't static. It's the same reason why, as an example, the MTE macerator's structure is done like that too. But not too sure on that one.

src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntity_QuantumForceTransformer.java
BlueWeabo: We are gonna be removing the catalyst housing itself as the catalyst won't be consumed. Pretty sure that was the plan

src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntity_QuantumForceTransformer.java
BlueWeabo: Will do it. I hate it when multis form without every min requirement.

src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntity_QuantumForceTransformer.java
BlueWeabo: Please go and migrate whole GT++. As I am pretty sure other multis also need to be migrated. Can do it later tho.

src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntity_QuantumForceTransformer.java
BlueWeabo: I forgot to remove it. Old things from my first iteration.

src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntity_QuantumForceTransformer.java
Glease: Really just lazy initialization vs eager initialization. First one save a bit  of memory and a bit of startup time, later make it not a final field which is against my personal taste.

src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntity_QuantumForceTransformer.java
```diff
@@ -0,0 +1,1160 @@
...
+    protected int mFabCoilTier = 0;
+    protected int mGlassTier = 0;
+    protected int mMinimumMufflerTier = 0;
+    private boolean mSeparateInputBusses = false;
+    private boolean mFluidMode = false, doFermium = false;
+    private Fluid mNeptunium = ELEMENT.getInstance().NEPTUNIUM.getPlasma();
+    private Fluid mFermium = ELEMENT.getInstance().FERMIUM.getPlasma();
+    private GT_MetaTileEntity_Hatch_Input mNeptuniumHatch;
+    private GT_MetaTileEntity_Hatch_Input mFermiumHatch;
+    private IStructureDefinition<GregtechMetaTileEntity_QuantumForceTransformer> STRUCTURE_DEFINITION = null;
```
Glease: OK I just noticed the field is not even static. This is just a waste of memory. This will not do. Make it a static field.

Glease (dismissed):

src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntity_QuantumForceTransformer.java
```diff
@@ -0,0 +1,1083 @@
...
+                            'H',
+                            ofChain(
+                                    buildHatchAdder(GregtechMetaTileEntity_QuantumForceTransformer.class)
+                                            .hatchClass(GT_MetaTileEntity_Hatch_Catalysts.class)
+                                            .adder(GregtechMetaTileEntity_QuantumForceTransformer::addCatalystHousing)
+                                            .casingIndex(TAE.getIndexFromPage(0, 10))
+                                            .dot(4)
+                                            .build(),
+                                    buildHatchAdder(GregtechMetaTileEntity_QuantumForceTransformer.class)
+                                            .atLeast(InputBus, InputHatch, Maintenance, Energy, ExoticEnergy)
```
minecraft7771: Use `Energy.or(ExocityEnergy)`

minecraft7771 (changes requested):

src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntity_QuantumForceTransformer.java
```diff
@@ -0,0 +1,990 @@
...
+                }
+
+                if (mCurrentMaxParallel >= mMaxParallel) {
+                    mCurrentMaxParallel = mMaxParallel;
+                    break;
+                }
+            }
+
+            if (mFermiumHatch != null) {
+                if (mFermiumHatch.getFluid().getFluid() == mFermium) {
```
minecraft7771: I'd prefer if you'd use the fluidEqual method over `==` for fluid comparision.

src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntity_QuantumForceTransformer.java
```diff
@@ -0,0 +1,990 @@
...
+            if (mMaxProgresstime == Integer.MAX_VALUE - 1 && mEUt == Integer.MAX_VALUE - 1) return false;
+
+            if (this.mEUt > 0) {
+                this.mEUt = (-this.mEUt);
+            }
+
+            int[] tChances;
+            if (aStack == null
+                    || aStack.getItemDamage() == 0
+                    || mNeptuniumHatch.getFluid().getFluid() != mNeptunium) {
```
minecraft7771: Same for here

minecraft7771 (dismissed):

BlueWeabo:
**Requires MUI branch to be merged in GT5u and GT++.**

_**Do not blindly merge >:(**_

Connor-Colenso:
I still think the QFT textures need some serious work, they just feel very odd in their current state and not really in line/style with anything else.
Screenshot 2022-12-13 at 18 49 18

boubou19:
agreed, the work made with the structure is kinda wasted by ugly textures.

Steelux8:
> I still think the QFT textures need some serious work, they just feel very odd in their current state and not really in line/style with anything else.

I stopped working on the textures after reaching this point, simply because textures can be changed at any time and the multi itself is more important. Just like every other ugly texture that the pack has, it can be changed later on.

Connor-Colenso:
It must be changed before merging otherwise it gets added to an endless todo list.

Steelux8:
After today's testing, I believe this is ready to be merged.

BlueWeabo:
Whoever can do a better texture for the force field please do. I do not believe in my skills

src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntity_QuantumForceTransformer.java
```diff
@@ -810,13 +810,15 @@ private boolean processRecipe(
                                     mat.getFluid(tRecipe.getOutput(i).stackSize * 1000);
                                 } else {
                                     Item aItem = tRecipe.getOutput(i).getItem();
+                                    int aDamage = tRecipe.getOutput(i).getItemDamage();
                                     int aStackSize = tRecipe.getOutput(i).stackSize;
-                                    tItemOutputs.add(new ItemStack(aItem, aStackSize * mCurrentParallel));
+                                    tItemOutputs.add(new ItemStack(aItem, aStackSize * mCurrentParallel, aDamage));
```
OneEyeMaker: What about `GT_Utility.copyAmount`?

BlueWeabo: GT_Utility.copyAmount, limits it to 64 in a stack still, but i realized another way i can do it. will push it

GlodBlock: use `GT_Utility.copyAmountUnsafe`

BlueWeabo: fixed i think

Dream-Master (approved):

PR #404, opened 2022-10-15
fix frame box and block cant localized

iouter:
it use %material now too.

and fluid in the cell can read the fluid key instead of adding more keys.

### DIFF
--- a/src/main/java/gtPlusPlus/core/block/base/BlockBaseModular.java
+++ b/src/main/java/gtPlusPlus/core/block/base/BlockBaseModular.java
@@ -28,6 +28,7 @@ public class BlockBaseModular extends BasicBlock {
     protected int blockColour;
     public BlockTypes thisBlock;
     protected String thisBlockMaterial;
+    protected String thisBlockMaterialTranslatedName;
     protected final String thisBlockType;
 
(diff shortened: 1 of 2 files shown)


eigenraven (changes requested):
Does this work if you switch languages while the game is running? You shouldn't keep translated names in block classes, they should only ever appear clientside. Also in some languages you can't just substitute the material name, because of declension. E.g. in polish: iron is "żelazo", while block of iron is "blok żelaza" (o -> a has to change), so there should be separate translations for the full names.

iouter:
> Does this work if you switch languages while the game is running?

gregtech.lang does not depend on the language, only on itself.

> You shouldn't keep translated names in block classes, they should only ever appear clientside.

could this cause any problems?

> Also in some languages you can't just substitute the material name, because of declension. E.g. in polish: iron is "żelazo", while block of iron is "blok żelaza" (o -> a has to change), so there should be separate translations for the full names.

i know russian also has this problem, maybe consider setting the automatic writing of %material through a config. i don't know if there can be a more general solution instead of having translators translate one by one.

eigenraven (approved):
I guess the problems are with GT's api and would have to be solved there, in that case I think this PR is fine as-is.

iouter:
which problem?

PR #415, opened 2022-11-01
New quantum anomaly texture

Connor-Colenso:
Full credit to Jimbno (as usual) for this texture. A much needed improvement.

q_anomaly


glowredman (approved):

POPlol333:
Textures are good, but isn't  it bad for ram if we keep adding moving texture to rams?

Glease:
128×128 is actually bigger than 16×384, so it's actually saving ram here

POPlol333:
oh it all depend from that? I though the fact that the texture was moving had a point in that

Glease:
Yeah animated takes up a bit more, but not quite as much. That ~~sprite abuse~~ unintended usage of sprite wouldn't have caused as much harm if it isn't a 400×400 res or if it was for two dozens items.

PR #418, opened 2022-11-04
Add tainted sapling to tgs

guineawheek:
It's a growable tree!


POPlol333 (approved):
Code-wise it's ok, but we might need feedback from magics fellow if it's op or no (I have no clue myself) @Alastors ?

PR #455, opened 2022-12-13
change how item color animation is ticked

### LINKED ISSUE #11958: Animated items flash faster when there are more of them

Incandescence~ ♡#8301

### Your Pack Version

2.2.7

### Your Server

SP

### Type of Server

None

### Your Expectation

I put myself in a situation where multiple animated GT++ textures were onscreen at the same time, and I thought that nothing special would happen.

### The Reality

It seems that all copies of the same GT++ animated item share an animation counter. See attached videos for a demonstration.
Hypogen (eyestrain warning): https://streamable.com/sp2h24
Celestial Tungsten (eyestrain warning): https://streamable.com/nxuro3
Chromatic Glass (worse eyestrain warning): https://streamable.com/hrv2ft
An entire Compressed Chest of Chromatic Glass (there was a point when this was about half full where this was terrible to look at, but the full thing isn't that bad because it's too fast for the monitor / the human eye to keep up; still, pretty bad for the eyes): https://streamable.com/1pt3df

### Your Proposal

Fix all animated items ticking the counter so that the speed doesn't increase

### Final Checklist

Glease:
Fix https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/11958

This changes it from receiving ticks from getColorFromItemStack (which will be called multiple times if there are multiple stacks on screen) to a function of system nanotime. Since minecraft client ticks are also based off system nanotime this probably wouldn't be a problem.


miozune (approved):
far saner implementation lol

PR #461, opened 2022-12-22
Fix QFT Parallel not calculated correctly

BlueWeabo:
GT_Utility.copyAmountUnsafe wasn't copying it correctly and applying the new stacksize.
Just multiplying the amount in the fluid stack wasn't enough apparently and broke if the qft had more than 1 parallel

### DIFF
--- a/src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntity_QuantumForceTransformer.java
+++ b/src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntity_QuantumForceTransformer.java
@@ -36,6 +36,7 @@ import net.minecraft.block.Block;
 import net.minecraft.client.renderer.RenderBlocks;
 import net.minecraft.client.renderer.Tessellator;
 import net.minecraft.entity.player.EntityPlayer;
+import net.minecraft.item.Item;
 import net.minecraft.item.ItemStack;
 import net.minecraft.nbt.NBTTagCompound;
 import net.minecraft.util.EnumChatFormatting;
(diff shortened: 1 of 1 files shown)


src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntity_QuantumForceTransformer.java
```diff
@@ -712,8 +712,7 @@ public boolean checkRecipe(final ItemStack aStack) {
 
     private boolean processRecipe(
             ItemStack[] aItemInputs, FluidStack[] aFluidInputs, GT_Recipe.GT_Recipe_Map aRecipeMap, ItemStack aStack) {
-        long tVoltage =
-                getMaxInputVoltage() / getExoticAndNormalEnergyHatchList().size();
+        long tVoltage = getMaxInputVoltage();
```
kuba6000: I think that should stay
With this change and 2 EV hatches its wrong:
(EV V + EV V) * (2 A + 2 A) != 4 A EV

BlueWeabo: if you look at how i calculate the amps, i am removing some amps. and afterwards in the OC function remove 0.05% eu/t from the total to get it to use less to account for "cable loss". GDCloud tested it to make sure it worked fine.

BlueWeabo: okay i see what you meant. will try to get a fix.

kuba6000: Its definitely gonna take too much power. Look at this example with 2 UV energy hatches
tVolatage = 524288 + 524288 = 1048576
tAmps = (2 + 2) * 0.8 ~= 3
total = 3145728
WHILE 2 UV hatches can take at most 2097152 EU/t

BlueWeabo: fixed

src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntity_QuantumForceTransformer.java
```diff
@@ -712,9 +712,10 @@ public boolean checkRecipe(final ItemStack aStack) {
 
     private boolean processRecipe(
             ItemStack[] aItemInputs, FluidStack[] aFluidInputs, GT_Recipe.GT_Recipe_Map aRecipeMap, ItemStack aStack) {
-        long tVoltage =
-                getMaxInputVoltage() / getExoticAndNormalEnergyHatchList().size();
-        long tAmps = (long) Math.floor(getMaxInputAmps() * 0.80);
+        int hatches = getExoticAndNormalEnergyHatchList().size();
+        long tVoltage = getMaxInputVoltage() / hatches;
+        // Need to check weather the hatches used are TT ones or not as TT hatches can request 20% more amps
+        long tAmps = (long) Math.floor(mExoticEnergyHatches.isEmpty() ? getMaxInputAmps() : getMaxInputAmps() * 0.80);
```
kuba6000, line 718: Wrong way around ? (it's the opposite way than the comment says or the comment should say less) . Also why would you cut off 20% of usable power ?

BlueWeabo: a 4A energy hatch from TT can accept 5A, which gets added to max Input Amps. a 64A hatch can accept up to 80A. (so technically it accepts 25% more and i want to remove it so i remove 20%)

kuba6000: Ah ok, I misunderstood the comment

BlueWeabo: it gets worse with lasers as a 256A laser is counted as a 320A hatch

kuba6000 (approved):

PR #470, opened 2023-01-02
Move NEI handlers to new recipemap methods

miozune:


repositories.gradle
```diff
@@ -1,6 +1,7 @@
 // Add any additional repositories for your dependencies here
 
 repositories {
+    mavenLocal()
```
Glease, line 4: ?

Glease (commented):
not sure about mavenLocal(), but otherwise is good to merge

repositories.gradle
miozune: Eh at least it doesn't cause crash on workflow. I'm also just lazy to type it when I test something

Glease: I'll just pretend it doesn't exist then. it's not like I don't have this locally..

Glease (approved):

PR #472, opened 2023-01-03
add a line in tricorder to say the mode of the multi machine

boubou19:
this is handy for those like me who have a non neglectible amount of Large Processing Factories and who forgot in what mode they were.

Now there is a line like this: 
image


draknyte1:
I believe there’s a method which isn’t final called extraInfo or something specifically for this..

POPlol333 (approved):

PR #481, opened 2023-01-06
Decreased minimum Mixer casings

RaitGamerGR:
With this change you can put all needed circuits, fluids into one mixer, if you dont need to build two


POPlol333 (approved):
code-wise I'm ok, not sure about balance

PR #486, opened 2023-01-08
Add batch mode to gt++ multis

GDCloudstrike:
Adds the batch mode functionality to gt++ multis. 
gt++ lag can get quite bad when processing fast recipes, batch mode can help reduce that lag

Here's a before and after comparison using a gt++ bending machine
image
image

### DIFF
--- a/src/main/java/gtPlusPlus/xmod/gregtech/api/metatileentity/implementations/base/GregtechMeta_MultiBlockBase.java
+++ b/src/main/java/gtPlusPlus/xmod/gregtech/api/metatileentity/implementations/base/GregtechMeta_MultiBlockBase.java
@@ -170,7 +170,7 @@ public abstract class GregtechMeta_MultiBlockBase<T extends GT_MetaTileEntity_Ex
     }
 
     protected float batchMultiplier = 1.0f;
-    protected int maxBatchSize = 128;
+    protected static final int MAX_BATCH_SIZE = 128;
 
     public abstract String getMachineType();
 
@@ -973,7 +973,7 @@ public abstract class GregtechMeta_MultiBlockBase<T extends GT_MetaTileEntity_Ex
         if (mUseMultiparallelMode) {
             int extraParallelRecipes = 0;
             for (;
-                    extraParallelRecipes + parallelRecipes < aMaxParallelRecipes * maxBatchSize;
+                    extraParallelRecipes + parallelRecipes < aMaxParallelRecipes * MAX_BATCH_SIZE;
                     extraParallelRecipes++) {
                 if (!tRecipe.isRecipeInputEqual(true, aFluidInputs, aItemInputs)) {
                     break;
@@ -1021,8 +1021,8 @@ public abstract class GregtechMeta_MultiBlockBase<T extends GT_MetaTileEntity_Ex
 
         this.mMaxProgresstime = Math.max(1, this.mMaxProgresstime);
 
-        if (mUseMultiparallelMode && mMaxProgresstime <= maxBatchSize / 2) {
-            mMaxProgresstime = maxBatchSize / 2;
+        if (mUseMultiparallelMode && mMaxProgresstime <= MAX_BATCH_SIZE) {
+            mMaxProgresstime = MAX_BATCH_SIZE;
         }
 
         // Collect fluid outputs
--- a/src/main/java/gtPlusPlus/xmod/gregtech/api/metatileentity/implementations/base/GregtechMeta_MultiBlockBase.java
+++ b/src/main/java/gtPlusPlus/xmod/gregtech/api/metatileentity/implementations/base/GregtechMeta_MultiBlockBase.java
@@ -941,7 +941,6 @@ public abstract class GregtechMeta_MultiBlockBase<T extends GT_MetaTileEntity_Ex
             return false;
         }
 
-
         // EU discount
         float tRecipeEUt = (tRecipe.mEUt * aEUPercent) / 100.0f;
         float tTotalEUt = 0.0f;
@@ -980,7 +979,6 @@ public abstract class GregtechMeta_MultiBlockBase<T extends GT_MetaTileEntity_Ex
             parallelRecipes += extraParallelRecipes;
         }
 
-
         // -- Try not to fail after this point - inputs have already been consumed! --
 
         // Convert speed bonus to duration multiplier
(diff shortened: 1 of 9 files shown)


src/main/java/gtPlusPlus/xmod/gregtech/api/metatileentity/implementations/base/GregtechMeta_MultiBlockBase.java
```diff
@@ -2452,6 +2465,23 @@ public boolean onSolderingToolRightClick(
         return true;
     }
 
+    public boolean mUseMultiparallelMode = false;
+
+    @Override
+    public boolean onWireCutterRightClick(
```
minecraft7771, line 2481: Wire cutter will collide with mixer, ABS and QFT.

src/main/java/gtPlusPlus/xmod/gregtech/api/metatileentity/implementations/base/GregtechMeta_MultiBlockBase.java
```diff
@@ -973,6 +979,11 @@ public boolean checkRecipeGeneric(
         // e.g. 100% speed bonus = 200% speed = 100%/200% = 50% recipe duration.
         aSpeedBonusPercent = Math.max(-99, aSpeedBonusPercent);
         float tTimeFactor = 100.0f / (100.0f + aSpeedBonusPercent);
+
+        if (mUseMultiparallelMode) {
```
minecraft7771: With the placement here you can overclock the batch itself, so you could theoretically 1 tick the batch of 128. Intended?

src/main/java/gtPlusPlus/xmod/gregtech/api/metatileentity/implementations/base/GregtechMeta_MultiBlockBase.java
GDCloudstrike: fixed for mixer and ABS, QFT batchmode has to be done separately, will make another pr and fix the wire cutter issue there

src/main/java/gtPlusPlus/xmod/gregtech/api/metatileentity/implementations/base/GregtechMeta_MultiBlockBase.java
GDCloudstrike: as it stands, yes

POPlol333:
oh lol workspace setup fucked up

src/main/java/gtPlusPlus/xmod/gregtech/api/metatileentity/implementations/base/GregtechMeta_MultiBlockBase.java
```diff
@@ -967,12 +967,28 @@ public boolean checkRecipeGeneric(
             return false;
         }
 
+        float batchMultiplier = 1.0f;
```
minecraft7771: The placement here will make batch-mode ignore void protection.

src/main/java/gtPlusPlus/xmod/gregtech/api/metatileentity/implementations/base/GregtechMeta_MultiBlockBase.java
```diff
@@ -967,12 +967,28 @@ public boolean checkRecipeGeneric(
...
+        if (mUseMultiparallelMode) {
+            int extraParallelRecipes = 0;
+            for (; extraParallelRecipes + parallelRecipes < aMaxParallelRecipes * 128; extraParallelRecipes++) {
```
minecraft7771: The 128 should be refactored into a constant (and best make it protected so you can use it in the inheriting multiblocks)

src/main/java/gtPlusPlus/xmod/gregtech/api/metatileentity/implementations/base/GregtechMeta_MultiBlockBase.java
```diff
@@ -2452,6 +2470,23 @@ public boolean onSolderingToolRightClick(
         return true;
     }
 
+    public boolean mUseMultiparallelMode = false;
```
minecraft7771: This should probably be `protected`

src/main/java/gtPlusPlus/xmod/gregtech/api/metatileentity/implementations/base/GregtechMeta_MultiBlockBase.java
minecraft7771: Also imo this should be done in the main parallel loop instead of being its own thing, since this loop ignores energy while the other respects if the machine has enough energy for so many parallels

minecraft7771 (changes requested):
My review of the one for-loop holds true in the other loops as well.

src/main/java/gtPlusPlus/xmod/gregtech/api/metatileentity/implementations/base/GregtechMeta_MultiBlockBase.java
minecraft7771: So by checking back with other devs it seems like this is unwanted behavior and so also needs addressing.

src/main/java/gtPlusPlus/xmod/gregtech/api/metatileentity/implementations/base/GregtechMeta_MultiBlockBase.java
minecraft7771: Nvmd blue reminded me thats now how batch mode works 💀

GDCloudstrike: yeah, the increase in recipe time already covers the power cost

src/main/java/gtPlusPlus/xmod/gregtech/api/metatileentity/implementations/base/GregtechMeta_MultiBlockBase.java
GDCloudstrike: That should be fine imo, people who use batch mode are very likely going to use an me output bus instead of normal buses

minecraft7771: If you want to continue that route it should atleast be explained to the user (eg adding it to the chat message when enabling batch mode).

GDCloudstrike: ok, will add that

minecraft7771 (approved):

PR #489, opened 2023-01-08
two fix -> Isamill

Quetz4l:
two fix from https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/12117

### DIFF
--- a/src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/processing/GregtechMetaTileEntity_IsaMill.java
+++ b/src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/processing/GregtechMetaTileEntity_IsaMill.java
@@ -256,14 +256,14 @@ public class GregtechMetaTileEntity_IsaMill extends GregtechMeta_MultiBlockBase<
                     } else {
                         if (aFoundEntity.getHealth() > 0) {
                             EntityUtils.doDamage(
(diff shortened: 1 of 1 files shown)


Glease (commented):
Just make the damagesource not bypass armor. No need to reinvent armor calculation yourself.

Quetz4l:
Isamill is like a big meat grinder and I think it's a good thing, if only really cool armor will stop all damage

Glease:
Here is the thing. I don't think armor point the correct scaling factor. Diamond set give 80% reduction, same as nano and quantum. Anything better is from magic (with shadow fortress giving 100%) or straight up infinity. It just feels so wrong to have diamond the same as nano and quantum. In fact, any armor better than gold (e.g. just iron) has 75% reduction. I don't think that is anyway close to "cool armor"...

However, I also don't know what should be the alternative, beyond whitelist each armor set (which is obviously very tedious!). Perhaps this can just make do. This is just a very minor thing after all...

Dream-Master:
@Zereff06 are you on Discord ?

Quetz4l:
> @Zereff06 are you on Discord ?

Yes, my username is Quetz4l or Кетцаль#0802

@Glease, I think the armor piercing mechanic is good in this case, according to the author's idea. I only reduced the damage inflicted when wearing armor, given the extra hearts, it will be hard to die.

Players who will build this will already be at a high tier and they just do not need to walk in front of a working machine. Removing the same damage makes the mechanics of damage almost meaningless.

If you still want to remove it, can you tell me which method to use for damage?

Glease (approved):
I'm convinced. This now have my approval.

PR #509, opened 2023-01-20
Add Sifting Recipe to Thorium for Radium

Steelux8:
This PR adds a Sifter recipe for Thorium similar to those for Uranium, Uraninite and Pitchblende, which output Radium-226 for the purpose of decaying and processing into Radon. This recipe has a lower chance for Radium outputs than any of the other 3.

The purpose of this change is to allow Radon to be more easily accessible before getting to t2 planets, which will become much harder with a change I'm planning to add LuV circuits to the rocket recipe.


Steelux8:
This PR is ready for review, and it also will be the main way of getting the first buckets of Radon if the tiny dust Radon loop recipe is removed.

Dream-Master (approved):

PR #516, opened 2023-01-23
Remove Australia dimension and associated world gen, mobs etc.

Connor-Colenso:
See title, tested in full pack. Loads fine.


boubou19:
what happens if previously generated australia is loaded with this GT++ version? nothing? a crash?

Dream-Master:
Australian don’t like this change :trollface:

Connor-Colenso:
> what happens if previously generated australia is loaded with this GT++ version? nothing? a crash?

I believe the dimension may load but any dim specific blocks will be deleted. I'm not 100% sure but given it's impossible to access this dimension without very lategame equipment (that was never intended to be used that way) it's not a big deal. Server owners can just delete the file if they really want to be 100% sure.

Steelux8 (approved):
I'm going to trust that this isn't going to cause problems on worlds where Australia was already generated, rare as they may be.

PR #527, opened 2023-02-02
Increase capacity and production rate of the reservoir hatch

GDCloudstrike:
Some chemical reactor recipe when done in a MCR would be able to use more than 100ML water every 5 seconds, the increase makes sure that the reservoir hatch always keeps up


OneEyeMaker:
Now reservoir hatch can be crafted at EV tier. But such "hungry" water consumers as MCR become available much later (LuV+). So what about adding several tiers of reservoir hatch (IV, LuV,...) with bigger capacity and water generation speed? (Instead of transforming EV hatch to OP creative water source).

BlueWeabo:
Don't we already get a water reservoir at EV, which just requires a single pump to get infinite water out? Its how we are able to keep up with the MCR's demands on water in our base.

OneEyeMaker (approved):

Dream-Master:
@GDCloudstrike is this only for one tier ?

GDCloudstrike:
It's unlocked at EV and there's only this one block. But likewise there is the reservoir that is unlocked at EV that is also infinite water, this just compresses the water input setup into a single block.

POPlol333:
I mean having infinite throughput at EV won't make your playthrought easier so it's fine

POPlol333 (approved):

PR #530, opened 2023-02-05
Cleanup recipe adder

miozune:


src/main/java/gtPlusPlus/xmod/gregtech/recipes/GregtechRecipeAdder.java
```diff
@@ -185,233 +101,49 @@ public boolean addCokeOvenRecipe(ItemStack aInput1, ItemStack aInput2, FluidStac
...
+                null,
+                new int[] {},
+                aFluidInput == null ? new FluidStack[] {} : new FluidStack[] { aFluidInput },
+                new FluidStack[] { aFluidOutput },
+                aDuration,
+                aEUt,
+                0);
+        GTPP_Recipe_Map.sMatterFab2Recipes.addRecipe(aRecipe);
+        RECIPEHANDLER_MatterFabricator.debug5(aFluidInput, aFluidOutput, aDuration, aEUt);
         return true;
```
OneEyeMaker, line 119: Please, make sure that return of every method is consistent. Right now some methods check change of size of recipe map, another ones just return true.

I think that returning true is much clearer behavior (instead of checking map size). Methods of RecipeAdder of GT also just return true.

OneEyeMaker (changes requested):
Good job!

There's just one thing that (I think) should be adjusted.

OneEyeMaker (approved):

PR #535, opened 2023-02-09
Fix immature alk code

Connor-Colenso:
Removes a bunch of swear words from the code and the class that breaks the open inventory achievement in dev.


miozune (approved):
yay

PR #549, opened 2023-02-19
Add a new option of generating power in early EV

27yearoldminecraftgaymer:
While active, the explosion generator consumes primed TNT or Industrial TNT in a 3x3 area behind its pusher plate to generate EU and a massive amount of pollution. The EU values are closely based off nitro-benzene. The crown has been worn for too long.

image

https://user-images.githubusercontent.com/125762480/219979723-2785746e-c3ca-4fb7-a3ae-6382c330d98d.mp4


chochem (changes requested):
That sounds potentially way to easy given the existence of the explosive bee.

We will need some actual numbers based on the good production methods.

This also sounds like a significant LV-IV change, so broader dev agreement will be needed.

POPlol333:
I like the idea, and if we'd need to vote between a new cool powergen and some stupid bee I'm all for the powergen

chochem:
even if you change the bee it doesnt change the fact that I want to see some hard numbers.

They say this is based on nitrobenzene. But it is available earlier and with an easier fuel. That sounds unbalanced. But who knows without hard numbers.

27yearoldminecraftgaymer:
> We will need some actual numbers based on the good production methods.

currently it's 5 million eu for industrial tnt and 2.5 million for regular tnt. the same production costs for nitrobenzene with a 170% rotor would generate ~1.5 million eu per industrial tnt, so this could be atleast halved.

as for the the pollution, this is not so much for balance but to provide a way to generate millions of pollution on demand.

chochem:
that sounds hilariously overtuned. :D

even assuming we change the bee! I guess I have to make some computations since you are not sharing yours:

first of all what turbine gives you 170% in EV? you can only do normal ones at that point. I think the best is probably manyuallan at 125%. So with 5m/1.6m/1.25=2.5 You are equating 1 itnt with 2500L of nitrobenzene.

an itnt is 100 nitric acid and 100 sulf acid + 4 toluene (either 400 toluene or 266 toluene+small amounts of sugar/pe)
you equate that to 2500L nitrobenzene which needs 
2500L nitric acid + 1500L sulf acid + 5000L dist water  + 2500L benzene (and around 1000 titanium intially cost for the chemplant)

And what is the best way to make toluene? Probably from heavy oil at a 1:1 ratio. Doesnt look very hard.

I think its fair to say you are **off by a factor 10-50** here at first glance.

Since Boubou likes a good metric: If somebody wants to do this properly I guess one needs to follow all these steps through to the end and count the full setup cost for a given eu/t target maybe only based on sulfur input and wood/heavy oil respectively.

27yearoldminecraftgaymer:
With a large shadow metal turbine (140%), 1000l of nitrobenzene would generate 2.24M EU. 
The eu value of 1000l of benzene is 504k, so the eu value of 1000l nitration mixture would be 868k.
Using wood tar to make toluene, it makes 1/4 as much as benzene.
Considering the solidifier recipe for making gelled toluene, the eu value of ITNT should be 979k.

However using heavy oil to make the toluene is much more efficient because you're producing 400/s with 2 EV DTs. Your only real bottleneck is nitric acid.

I think EU values of 600k and 400k are reasonable to keep this competitive through later tiers. It might need a more expensive recipe/structure and lubricant consumption to keep it in line.

src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntity_ExplosionGenerator.java
```diff
@@ -0,0 +1,288 @@
...
+        if (mMachine) return -1;
+        return survivialBuildPiece(mName, stackSize, 1, 1, 0, elementBudget, env, false, true);
+    }
+
+    @Override
+    public boolean checkMachine(IGregTechTileEntity aBaseMetaTileEntity, ItemStack aStack) {
+        mCasing = 0;
+        mDynamoHatches.clear();
+        mMaintenanceHatches.clear();
+        return checkPiece(mName, 1, 1, 0) && mCasing >= 6 && checkHatch() && mDynamoHatches.size() <= 1;
```
Glease, line 232: You probably want it to require exactly one dynamo instead of 0 or 1... just to make sure less people are confused.

src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntity_ExplosionGenerator.java
```diff
@@ -0,0 +1,288 @@
...
+                            x -> x.setPosition(aX - offsetX, aY - offsetY, aZ - offsetZ).run());
+        }
+    }
+
+    @Override
+    public long maxEUStore() {
+        return 10000000;
+    }
+
+    private long addEnergyToHatch(MetaTileEntity aHatch) {
```
Glease, line 198: There is addEnergyOutput in base class. Why reinvent the wheel?

src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntity_ExplosionGenerator.java
```diff
@@ -0,0 +1,288 @@
...
+        return checkPiece(mName, 1, 1, 0) && mCasing >= 6 && checkHatch() && mDynamoHatches.size() <= 1;
+    }
+
+    @Override
+    public IMetaTileEntity newMetaEntity(IGregTechTileEntity aTileEntity) {
+        return new GregtechMetaTileEntity_ExplosionGenerator(this.mName);
+    }
+
+    @Override
+    public void saveNBTData(NBTTagCompound aNBT) {
```
Glease, line 241: There two NBT related method don't do anything except calling super. should be removed.

src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntity_ExplosionGenerator.java
```diff
@@ -0,0 +1,288 @@
...
+        List<Entity> entities = baseTile.getWorld().getEntitiesWithinAABBExcludingEntity(
+                null,
+                AxisAlignedBB.getBoundingBox(
+                        baseTile.getXCoord() - 0.5 - offsetX,
+                        baseTile.getYCoord() - 0.5 - offsetY,
+                        baseTile.getZCoord() - 0.5 - offsetZ,
+                        baseTile.getXCoord() + 1.5 - offsetX,
+                        baseTile.getYCoord() + 1.5 - offsetY,
+                        baseTile.getZCoord() + 1.5 - offsetZ));
+        for (Entity tnt : entities) {
```
Glease, line 147: I would suggest listening for `net.minecraft.world.Explosion` event. Running a get entity within aabb call per tick is slightly concerning to me, performance wise. would also remove the need for sending sound events manually

Glease (commented):
This doesn't consider the balance implication yet.

chochem:
> With a large shadow metal turbine (140%), 1000l of nitrobenzene would generate 2.24M EU. The eu value of 1000l of benzene is 504k, so the eu value of 1000l nitration mixture would be 868k. Using wood tar to make toluene, it makes 1/4 as much as benzene. Considering the solidifier recipe for making gelled toluene, the eu value of ITNT should be 979k.
> 
> However using heavy oil to make the toluene is much more efficient because you're producing 400/s with 2 EV DTs. Your only real bottleneck is nitric acid.
> 
> I think EU values of 600k and 400k are reasonable to keep this competitive through later tiers. It might need a more expensive recipe/structure and lubricant consumption to keep it in line.

Nitration mixture does not have an eu value, no. Thats not how that works, you can't get an eu value by substraction. And the eu value of benzene does not play into this at all from what I can tell.

And oh, I think I spot a problem in your calculation though, you seem to think itnt takes 1000L nitration mixture. It doesnt.

Also large turbines are not available in EV, that is IV. (or maybe very late EV depending on your definition) but with large you can get 150% and then later in IV more.

So yea, the balance question still needs someone serious work and more serious calculations.

Steelux8:
As a GT++ and powergen dev, I really do want to see detailed maths on optimal power generation with TNT here.

src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntity_ExplosionGenerator.java
27yearoldminecraftgaymer: looks like adding 800k energy output at once blows up the dynamo so I need that to buffer it.

src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntity_ExplosionGenerator.java
27yearoldminecraftgaymer: yeah I agree a machine idling at 10k ns isn't ideal but wouldn't the explosion have to happen to listen for it? this is taken from the botania flower that calls aabb in a 24x24 area every tick.

Dream-Master:
@27yearoldminecraftgaymer @chochem any news on this ? If the bee is the problem we change the bee not the new power gen but sure the math behind it need to fit into GT tiering and balancing

27yearoldminecraftgaymer:
>@27yearoldminecraftgaymer @chochem any news on this ? If the bee is the problem we change the bee not the new power gen but sure the math behind it need to fit into GT tiering and balancing

I think at the very least you could normalize the explosive bee and give it a comb that requires nitric acid to make the tnt. I wouldn't balance something around something I pred as a meme.

The structure should be more expensive (replace the solid steel and gearboxes with tungstensteel?), and the 400k/600k power values listed earlier seem reasonable to me. If you decide on what's balanced I can make some changes.

Steelux8:
We have enough powergen options as it is in early-game. This one has a very basic structure and little to no creativity or potential in how it generates power. I do not approve of it as it is, at least wait until the new MuTE system is finished and try to make this multi by using that system, with new mechanics to it.

Dream-Master:
@27yearoldminecraftgaymer are you still work and improve this pr or should we close this ?

Steelux8:
Due to the lack of improvements, the amount of merge conflicts that have piled up and the lack of concept of the original idea, I'm going to close this PR. It's been months and there is still nothing that shows a direct TNT powergen is worth adding.

PR #562, opened 2023-03-09
Fix removed CustomName init call causing an NPE

eigenraven:
The call was removed together with the gt5u enum extensions, but this one targets ic2 and must still run


Dream-Master (approved):

Glease:
guess I removed it by accident.. thanks for the fix!

PR #564, opened 2023-03-12
set the amount of muffler hatches to 1 in the MABS

boubou19:
Similar changes to https://github.com/GTNewHorizons/bartworks/pull/299 but for MABS.

I observed @Sphyix 's first MABS to take 90 µs just with the muffler hatches. So with 6 of them on a server, it's already more than 0.5ms of tick time spent on something useless.


BlueWeabo (approved):
muffler spam = 0. hopefully

boubou19:
hopefully

PR #565, opened 2023-03-12
follow up to https://github.com/GTNewHorizons/GTplusplus/pull/564

boubou19:
forgot to change the tooltip


BlueWeabo (approved):
at least we won't have a multi that never forms. and correct documentation

PR #574, opened 2023-03-17
Remove For Loops on Speed Multiplier for Multiplication

BlueWeabo:
Remove a not needed for loop in favor of multiplication.


minecraft7771 (approved):
Thanks for removing the commented out code and logging too.

PR #587, opened 2023-04-01
Update Rocketdyne Tooltip with More Info

Steelux8:
After building the Rocketdyne (Rocket Fuel Engine power generator) multi in survival, I noticed some things that were not documented, including previous changes of mine. I added more useful information to the tooltip.

### DIFF
--- a/src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntity_LargeRocketEngine.java
+++ b/src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntity_LargeRocketEngine.java
@@ -89,9 +89,12 @@ public class GregtechMetaTileEntity_LargeRocketEngine extends
     protected GT_Multiblock_Tooltip_Builder createTooltip() {
         GT_Multiblock_Tooltip_Builder tt = new GT_Multiblock_Tooltip_Builder();
         tt.addMachineType(getMachineType()).addInfo("Controller Block for the Large Rocket Engine")
+                .addInfo("Generating Power from Rocket Fuels - Supports TecTech Multi-Amp Dynamos!")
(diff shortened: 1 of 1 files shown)


src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntity_LargeRocketEngine.java
```diff
@@ -89,9 +89,12 @@ public GregtechMetaTileEntity_LargeRocketEngine(final String aName) {
     protected GT_Multiblock_Tooltip_Builder createTooltip() {
         GT_Multiblock_Tooltip_Builder tt = new GT_Multiblock_Tooltip_Builder();
         tt.addMachineType(getMachineType()).addInfo("Controller Block for the Large Rocket Engine")
+                .addInfo("Generating Power from Rocket Fuels - Supports TecTech Multi-Amp Dynamos!")
                 .addInfo("Supply GT++ Rocket Fuels and 1000L of " + mLubricantName + " per hour")
                 .addInfo("Produces as much energy as you put fuel in, with optional boosting")
+                .addInfo("This multi sometimes doesn't accept fluids if not enabled - enable it first!")
```
wlhlm: "sometimes" is phrased a bit awkwardly, this makes me think like is there some random chance this multi won't accept fluids.

We can be more precise here, something like "Fluid input can be toggled in the GUI", (I actually don't know if it's in the GUI as I'm unfamiliar with this multi, but I hope you get the idea), or "If this multi doesn't accept fluids, make sure input is enabled".

Steelux8: I'm not familiar myself, I found out for the first time in survival that it just wasn't accepting fluids, but it changed when I enabled the multi. I don't really know what's going on there.

Dream-Master (approved):

PR #591, opened 2023-04-05
Add tier tooltips to Charge Pack and quad batteries

miozune:


Dream-Master (approved):

boubou19:
@miozune can you do the PR on my branch instead pls? i have fixed the issue as to why it doesn't load.

miozune:
Doesn't it squash this change when merged to master? Pushing separated fix to master and rebasing current 2 PRs would be better I'd guess

boubou19:
if you know how to do it sure, my git knowledge is too limited to do this kind of branch manipulation.

the fix is this class src/main/java/gtPlusPlus/xmod/gregtech/loaders/RecipeGen_BlastSmelterGT_GTNH.java in 
https://github.com/GTNewHorizons/GTplusplus/commit/0073f79778e52fedbb1bffc9461582b826a9ae29

but because that's not the only class edited in the commit, i would think one has to actually make a new commit with the same code edit. I guess i'll be able to fix any merge conflict.

PR #598, opened 2023-04-11
fix incorrect ulv turbine tooltip

### LINKED ISSUE #13186: Very low indicated efficiency on Simple Steam Turbine

immibis#3179

### Your Pack Version

2.2.3

### Your Server

private server

### Java Version

Java 8

### Type of Server

Don't know

### Your Expectation

While looking for something in NEI I happened to notice the recipe for the "Simple Steam Turbine" and I expected some reasonable values.

### The Reality

Tooltip for the "Simple Steam Turbine" says 7% fuel efficiency. I did **not** test the actual fuel efficiency in-game.

### Your Proposal

It was probably meant to be 97%, following the general trend of decreasing fuel efficiency with higher voltage. Maybe it was meant to be 70%. Maybe the block does have 97% or 70% efficiency and only the tooltip is wrong, in which case the tooltip should be corrected.

### Final Checklist

Glease:
fix https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/13186


Dream-Master (approved):

immibis:
for anyone confused why `getEfficiency` is not edited: the [base class](https://github.com/GTNewHorizons/GT5-Unofficial/blob/master/src/main/java/gregtech/common/tileentities/generators/GT_MetaTileEntity_SteamTurbine.java#L60) calculates the efficiency as (600 / efficiency)%. Not sure how the generation is calculated, but that's how the tooltip is calculated. The simple formula of efficiency% which was being overridden here was incorrect.

PR #600, opened 2023-04-11
Fix migration code for ABS NBT

minecraft7771:
As discussed in https://github.com/GTNewHorizons/GT5-Unofficial/pull/1862, this is the correct solution for migrating the NBT.


BlueWeabo (approved):
better later than never

PR #604, opened 2023-04-17
Don't require a min of output busses on sifter

minecraft7771:
Unnecessary bloat, since ME output bus exists.


LewisSaber (changes requested):
Change tooltip too

LewisSaber (approved):

PR #605, opened 2023-04-20
Fix Material

Dream-Master:
Need this pr

* https://github.com/GTNewHorizons/GT5-Unofficial/pull/1893


Phineasor (approved):
Needed for Steller cat Pr. All of this works in main pack, just changing UEV+ mats to the new class. So many to go through. Hopefully no UEV mat references were missed.

PR #608, opened 2023-04-25
ForgeDirection 

mitchej123:
byte -> ForgeDirection


mitchej123:
A few errors like this... hard coded bits
```
/home/jason/dev/mc/GTplusplus/src/main/java/gtPlusPlus/xmod/gregtech/common/render/GTPP_Render_MachineBlock.java:63: error: incompatible types: int cannot be converted to ForgeDirection
                            getTexture(tMetaTileEntity, ForgeDirection.DOWN, 9, -1, false, false),
                                                                             ^
``

leagris (approved):

PR #618, opened 2023-05-05
Add channels to MABS

kstvr32:
https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/13241
Adds channels to MABS for coils and glass


BlueWeabo (approved):
someone actually bothered? nice good job

PR #628, opened 2023-05-14
Rotatable Wash Plant

MeiTianyou:
This allows the wash plant to be rotated. I think it looks pretty cool from the side if you cover it with glass.
wash_plant

I think it can look good using the flowing water for effect as well

rotate_for_aesthetics

### DIFF
--- a/src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/processing/GregtechMetaTileEntity_IndustrialWashPlant.java
+++ b/src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/processing/GregtechMetaTileEntity_IndustrialWashPlant.java
@@ -25,9 +25,11 @@ import net.minecraftforge.fluids.FluidStack;
 
 import com.gtnewhorizon.structurelib.alignment.IAlignmentLimits;
 import com.gtnewhorizon.structurelib.alignment.constructable.ISurvivalConstructable;
+import com.gtnewhorizon.structurelib.alignment.enumerable.ExtendedFacing;
 import com.gtnewhorizon.structurelib.structure.IStructureDefinition;
 import com.gtnewhorizon.structurelib.structure.ISurvivalBuildEnvironment;
 import com.gtnewhorizon.structurelib.structure.StructureDefinition;
+import com.gtnewhorizon.structurelib.util.Vec3Impl;
 
 import gregtech.api.enums.TAE;
 import gregtech.api.interfaces.IIconContainer;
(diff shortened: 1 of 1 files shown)


src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/processing/GregtechMetaTileEntity_IndustrialWashPlant.java
```diff
@@ -138,8 +140,7 @@ public boolean checkMachine(IGregTechTileEntity aBaseMetaTileEntity, ItemStack a
 
     @Override
     protected IAlignmentLimits getInitialAlignmentLimits() {
-        // don't rotate a washer, water will flow out.
```
boubou19, line 141: this is exactly why it was not made rotatable in the first place.

boubou19 (changes requested):

src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/processing/GregtechMetaTileEntity_IndustrialWashPlant.java
MeiTianyou: You can block off the "top" to stop the water flowing out. Besides that, letting the water flow out is also kind of cool looking, if you build a catchment for it.

boubou19: if you need extra handling when the multiblock is rotated, that's means it shouldn't being rotated.

It makes no sense that the washer should be rotated, as there is nothing happening in the water, you see nothing.

MeiTianyou: The extra handling to fill up the washer is there regardless of this change.

boubou19: did you know the washer is filling itself up?

MeiTianyou: Yes, of course. That's what the changes in checkForWater are for, they make sure the multi fills itself up correctly even when rotated

boubou19:
looks like Dream find this weird too, closing this.

MeiTianyou:
Sad but OK, feel free to pull the checkForWater() refactor by itself if you like. I think it's a little clearer as to what's going on than the original.

PR #635, opened 2023-05-21
Fix HV semifluid gen recipe

### LINKED ISSUE #13002: HV Semi-fluid generator recipe requires IV ABS for Inconel-625

Admiral#1878

### Your Pack Version

2.3.0

### Your Server

SP

### Java Version

Java 19

### Type of Server

Single Player

### Your Expectation

I expect to be able to craft the generators of the tier while in the tier.  I can craft an MV semi-fluid generator in MV, and I can craft an EV semi-fluid generator in EV.  The quests also encourage you to craft a semi-fluid generator to power your distillation tower, so it definitely expects you to be able to do so.

### The Reality

The Inconel-625 gears (which are part of the recipe) are not craftable without using the ABS.  The dust can be smelted in the EBF, but there is no mixer recipe for the dust (unlike the gears required for EV).

### Your Proposal

Either Inconel-625 dust should be craftable in a tier-appropriate (HV or below) mixer, or a different material should be chosen for the gears in the recipe.

### Final Checklist

chochem:
fixes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/13002


miozune:
How about making Inconel-625 craftable with mixer? There're still some items that require it for HV, e.g. auto chisel, charge pack, energy buffer

chochem:
there is already way too much in the mixer now. mixer is boring

Dream-Master (approved):

Dream-Master (approved):

PR #645, opened 2023-05-30
Reservoir hatch refill fix

GDCloudstrike:
Changed the refill amount of the reservoir hatch back to half of its max capacity every 5 seconds to avoid it only refilling when at 0 capacity, potentially causing machines to shut down


miozune (approved):
oh that's what you intended, I see

PR #649, opened 2023-05-31
Nerf low-grade plasmas in the XL Plasma turbine

S4mpsa:
Low-tier plasmas (primarily Helium) produce less EU/t for the same amount of fuel if the turbine EU/t is above a limit based on the fuel value of plasmas, following the formula  $\text{EU/t}=\text{EU/t}*\text{Min}(1.0, (\frac{ \text{fuelValue}  }{100})^2 / \text{EUPerTurbine})$

Some example values for various turbines:
image

### DIFF
--- a/src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/turbines/GT_MTE_LargeTurbine_Plasma.java
+++ b/src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/turbines/GT_MTE_LargeTurbine_Plasma.java
@@ -8,7 +8,6 @@ import net.minecraftforge.fluids.Fluid;
 import net.minecraftforge.fluids.FluidRegistry;
 import net.minecraftforge.fluids.FluidStack;
 
-import gregtech.api.enums.Materials;
 import gregtech.api.interfaces.ITexture;
 import gregtech.api.interfaces.metatileentity.IMetaTileEntity;
 import gregtech.api.interfaces.tileentity.IGregTechTileEntity;
@@ -24,10 +23,6 @@ import gtPlusPlus.xmod.gregtech.api.metatileentity.implementations.GT_MetaTileEn
 public class GT_MTE_LargeTurbine_Plasma extends GregtechMetaTileEntity_LargerTurbineBase {
 
     private static final HashSet<Fluid> BLACKLIST = new HashSet<>();
-
-    static {
-        BLACKLIST.add(Materials.Helium.getPlasma(0).getFluid());
-    }
 
     public GT_MTE_LargeTurbine_Plasma(int aID, String aName, String aNameRegional) {
         super(aID, aName, aNameRegional);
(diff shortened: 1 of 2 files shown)


src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/turbines/GT_MTE_LargeTurbine_Plasma.java
```diff
@@ -129,8 +129,15 @@ public boolean checkRecipeGeneric(ItemStack[] aItemInputs, FluidStack[] aFluidIn
                                     Float.MIN_NORMAL,
                                     ((GT_MetaGenerated_Tool) aStack.getItem()).getToolStats(aStack).getSpeedMultiplier()
                                             * GT_MetaGenerated_Tool.getPrimaryMaterial(aStack).mToolSpeed
-                                            * 50))
-                            * getSpeedMultiplier();
+                                            * 50));
+
+                        //Calculate total EU/t (as shown on turbine tooltip (Fast mode doesn't affect))
+                    double aEUPerTurbine = aTotalOptimalFlow
+                                    * 40 * 0.0105
```
boubou19: maybe it's classical info if you toy with plasma gen calculus every day, but from where the `40*0.0105` comes from?

src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/turbines/GT_MTE_LargeTurbine_Plasma.java
```diff
@@ -129,8 +129,15 @@ public boolean checkRecipeGeneric(ItemStack[] aItemInputs, FluidStack[] aFluidIn
...
+                                    * GT_MetaGenerated_Tool.getPrimaryMaterial(aStack).mPlasmaMultiplier
+                                    * (50.0f + (10.0f * ((GT_MetaGenerated_Tool) aStack.getItem()).getToolCombatDamage(aStack)));
```
boubou19: This line is ugly and should be extracted to a variable for better readability imo

src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/turbines/GT_MTE_LargeTurbine_Plasma.java
S4mpsa: The fuel value (as shown on the tooltip) can be seen [here](https://github.com/GTNewHorizons/GT5-Unofficial/blob/f2648a0eee49982e9fddbf1d0367530e3861526c/src/main/java/gregtech/api/items/GT_MetaGenerated_Tool.java#L513), I just removed the division and absorbed the factor of 50 from the above line, so 2000 -> 40.

boubou19:
Overall looks good to see some plasma getting nerfed. But i guess it's only half the work,  as it seems the plasma values are way off for some of them.

src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/turbines/GT_MTE_LargeTurbine_Plasma.java
boubou19: and for the `0.0105` ? tuning value for the nerfs?

S4mpsa: Magic value extracted from the value on tooltip, I just removed the divison by 100.

boubou19 (approved):

PR #664, opened 2023-06-08
Update dep + change comb recipe

Runakai1:
image
Now that QFT recipes are in the PlatinumSludge exception list, this is the proper recipe

updated bartwork dep to 0.7.17


BlueWeabo (approved):
bees? bees

Steelux8:
image

I bothered Runa to add a fluid input because I don't like the QFT having a recipe that is all items into all fluids. As compensation, the recipe now has 4x inputs and 4x outputs compared to normal platline skip, instead of only 2x outputs.

Steelux8 (approved):

PR #667, opened 2023-06-13
Remove Dragonblood from comb qft recipe

Runakai1:
This removes the unreal 100l of dragonblood in QFT.


BlueWeabo (approved):
Were you trying to kill thousands of generations of dragons?

PR #651, opened 2023-06-03
Add New Recipes for Powergen Plasmas

Steelux8:
This PR comes as a follow-up to the one that changed XL Plasma Turbines to apply penalties to plasmas based on their fuel value. The idea is that Helium and Tin plasmas won't be used from ZPM to many tiers later, upscaled indefinitely with compacts and XLs.

As an alternative, I suggest these fusion recipes for plasmas that already exist in the pack, generated by GT++ (which is why this PR is in GT++), but which have no recipes. This makes it so Plasma Mixer setups are unchanged. The fusion recipes follow a chain, instead of being just two common materials fused into a plasma to be used for power directly:

Mk1: Boron Plasma + Calcium Plasma -> Neon Plasma
Neon Plasma +  Molten Bedrockium -> Force Plasma

Mk2: Niobium Plasma + Zinc Plasma -> Krypton Plasma
Krypton Plasma + Force Plasma -> Astral Titanium Plasma
Astral Titanium Plasma +  Molten Titansteel -> Runite Plasma

Mk3: Molten Curium + Americium Plasma -> Xenon Plasma
Xenon Plasma + Runite Plasma -> Advanced Nitinol Plasma
Advanced Nitinol Plasma + Molten Tartarite -> Celestial Tungsten Plasma

I wrote a txt file with more information, and pasted it inside the spoiler below.

<details>
  <summary>Full doc</summary>
    Fusion chain logic: two fusion reactor outputs fuse into a new plasma, and then a material of the appropriate GT tier is fused into that plasma to create the power-dense plasma. At mk2 and mk3, the previous plasma needs to be added to the chain to make the new one.
  
  The first fusion makes a noble gas plasma. For mk2 and mk3, the next step makes a "magical metal" plasma by fusing the noble gas with the previous tier's complete plasma. The last step is to add a GT tier-appropriate material, which makes the power-dense plasma - Force, Runite or Dragonblood.
  
  All the new plasmas already exist in the pack, created by GT++. They just need to receive recipes.
  
  Mk1: Boron Plasma + Calcium Plasma -> X Plasma
  X Plasma +  Molten Bedrockium -> Improved X Plasma
  
  Mk2: Niobium Plasma + Zinc Plasma -> Y Plasma
  Y Plasma + Improved X Plasma -> XY Plasma
  XY Plasma +  Molten Titansteel -> Improved XY Plasma
  
  Mk3: Molten Curium + Americium Plasma -> Z Plasma
[...]

PR #669, opened 2023-06-13
Make the New Plasma Powergen Chains Viable

Steelux8:
See #651 for the PR that introduced these plasma chains. Big thanks to @S4mpsa for making the spreadsheet and helping me decide how these chains should be balanced to be useful, yet not too strong.

[PlasmaUnfuckening.xlsx](https://github.com/GTNewHorizons/GTplusplus/files/11738014/PlasmaUnfuckening.xlsx)

This is the spreadsheet Sampsa made, with a general overview of the numbers involved in this PR.

## What's the objective?

These plasma chains are meant to replace the Helium/Tin Plasma spam that has since been heavily nerfed when used together with XL Plasma Turbines, which was the usual setup. The final powergen numbers are somewhat similar to what Helium Plasma could do with compacts before, but the setup is a lot more extensive and more materials are needed to automate the chain. Hopefully, this amount of power will be enough to sustain a base until Dyson, but it won't be anywhere as simple as how fusion has been until the nerfs.

## What are the fusion chains like?

There are three chains, one for mk1 reactors, another for mk2 and the last for mk3. Each chain needs the final plasma of the previous one to be made. The focal point for compact reactors and XL turbines is the mk3 chain, which is the one that can match the previous powergen's numbers, but with a lot more complexity. The other two chains aren't as relevant for this stage, but they are both more powerful per fusion reactor than the 1-recipe setups of Helium and Tin Plasma, with an added cost like everything else.

## What would a compact reactor setup look like?

The final recipe that makes the strongest plasma, Celestial Tungsten Plasma, is the slowest recipe, so it's the only one that _needs_ to be a compact. However, other recipes might also be moved to compacts because they might only be 6x as fast, which would force a number of normal reactors. The entire mk3 chain, which includes the other two, starts from Helium Plasma and requires 14 new fusion steps to be completed, but the power per reactor is considerably higher if there are no bottlenecks before the final recipe. The complete setup would require a good number of reactors either way, but the added complexity and the choice of which chain to build between the three should add more entertainment and choice possibilities to the player, rather than simply choosing between Helium or Tin Plasma spam. 

## Other notes:

Celestial Tungsten Plasma's fuel value was buffed to 720M in order to improve the EU per reactor by more than 4x the mk2 equivalent, to account for the 4x difference in EU per amp across tiers. Additionally, this new fuel value avoids penalties even in the spacetime turbines.

### DIFF
--- a/src/main/java/gtPlusPlus/xmod/gregtech/loaders/RecipeGen_Plasma.java
+++ b/src/main/java/gtPlusPlus/xmod/gregtech/loaders/RecipeGen_Plasma.java
@@ -46,7 +46,7 @@ public class RecipeGen_Plasma extends RecipeGen_Base {
                     case "Runite":
                         GT_Values.RA.addFuel(GT_Utility.copyAmount(1L, aPlasmaCell), aContainerItem, 350_000, 4);
                     case "CelestialTungsten":
-                        GT_Values.RA.addFuel(GT_Utility.copyAmount(1L, aPlasmaCell), aContainerItem, 600_000, 4);
+                        GT_Values.RA.addFuel(GT_Utility.copyAmount(1L, aPlasmaCell), aContainerItem, 720_000, 4);
                     default:
                         GT_Values.RA.addFuel(
                                 GT_Utility.copyAmount(1L, aPlasmaCell),
--- a/src/main/java/gtPlusPlus/xmod/gregtech/loaders/recipe/RecipeLoader_Nuclear.java
+++ b/src/main/java/gtPlusPlus/xmod/gregtech/loaders/recipe/RecipeLoader_Nuclear.java
@@ -497,22 +497,22 @@ public class RecipeLoader_Nuclear {
 
         // Mk2
         GT_Values.RA.stdBuilder().noItemInputs().noItemOutputs()
-                .fluidInputs(Materials.Niobium.getPlasma(100), Materials.Zinc.getPlasma(100))
-                .fluidOutputs(new FluidStack(ELEMENT.getInstance().KRYPTON.getPlasma(), 100)).duration(3 * SECONDS)
+                .fluidInputs(Materials.Niobium.getPlasma(144), Materials.Zinc.getPlasma(144))
+                .fluidOutputs(new FluidStack(ELEMENT.getInstance().KRYPTON.getPlasma(), 100)).duration(32 * TICKS)
                 .eut(TierEU.RECIPE_ZPM).metadata(FUSION_THRESHOLD, 300000000).addTo(sFusionRecipes);
 
         GT_Values.RA.stdBuilder().noItemInputs().noItemOutputs()
                 .fluidInputs(
-                        new FluidStack(ELEMENT.getInstance().KRYPTON.getPlasma(), 100),
-                        new FluidStack(ELEMENT.STANDALONE.FORCE.getPlasma(), 100))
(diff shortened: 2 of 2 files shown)


kstvr32:
Based on the spreadsheet values, I like this change a lot. It keeps a similar EU/t per reactor but adds complexity and new ways to gain PLE materials. 👍

S4mpsa (approved):
Looks good

fluffle:
Do you plan to balance the EU/t available via the Extreme HX -> XL SC turbine route in a similar fashion? It doesn't look like this PR makes any changes to the quantities of SC steam produced by these plasmas, and some back-of-the-envelope maths -- sorry, i don't have a nice spreadsheet like @S4mpsa ;-) -- shows that there is very little value in going past Runite, or even Astral Titanium, when going that route.

For example, to replace my 2x T3 fusion kicking out tin plasma to generate 6,144,000 L/s SC steam for an XL turbine with 16 cosmic neutronium rotors, I need 720L/s of Astral Titanium (roughly 0.57 T3 reactors), 585L/s of Runite (0.47 T3) or 570L/s Celestial Tungsten (0.45 T3). Given the huge jump in plasma energy from Runite to Celestial Tungsten, it seems like something is off here. The XL SC turbine will kick out ~75.2M EU/t or ~1.5B EU/s so we can calculate the efficiency multiplier of this path v.s. a an XL plasma turbine. (2.45 is rotor efficiency)

720L Astral Titanium = 204.96M EU * 2.45 = 502M so efficiency multiplier of SC route = ~3x
585L Runite = 204.75M EU * 2.45 = 501M = ~3x again
570L Celestial Tungsten = 410.4M EU = 1B = ~1.5x

 Assuming the efficiency gain from runite / astral titanium through the extreme HX is the desired balance, the HX should need ~285L/s of CT plasma. Or, 1500L of CT plasma should kick out ~32.4M L of SC steam per second instead of 16.2M L. Or (argh) you could nerf the efficiency gain of the HX to make CT plasma look better.

More directly:

1500L CT plasma = 16.2M L SC = 1.08B EU = 66.66 EU/L
1500L Runite plasma = 15.75M L SC = 33.33EU/L
1500L AT plasma = 12.8M L SC = 33.33EU/L
1500L Nb plasma = 12.13M L SC = 404.3M EU = 33.33 EU/L
1500L Tin plasma = 4.5M L SC = 50EU/L

So I assume that 33.33 EU/L was the balance point previously and these changes in EU/L just need to be reflected in the SC numbers?

fluffle:
Ah, I found why celestial tungsten is getting limited, at least:

https://github.com/GTNewHorizons/GoodGenerator/blob/f5f68259301e49927877b442fe51831a651cde25/src/main/java/goodgenerator/loader/RecipeLoader_02.java#L1631C28-L1631C28

720,000 * 1500 * 160 = 3.24B which is > 2^31

But plasmas don't ever create superheated steam, they always output supercritical. So perhaps we could set the SH steam value to 0 and remove this check? If that sounds OK I can send a PR for it.

That code also explains the weird tin numbers -- I missed that tin uses 1000 plasma to generate 4.5M L SC steam, not 1500. So that's fine, it's just celestial tungsten that's broken.

Steelux8:
I have not tested the plasma output through SC turbines, because those have been problematic for a while and have multiple bugs that could lead to even more ridiculous plasma setups. The more people use them, and the more plasma-related content is added, the more issues I see coming up, and I honestly have never found how to properly support these turbines.

That said, the PRs related to this have been approved. Maybe complete fixes will be made in time.

PR #677, opened 2023-06-22
add solidifier hatches

Quetz4l:
New hatches for Large Processing Factory. Without these hatches you have to make one multiblock per mold. Also specialized hatch add you a 50% bonus to your speed (250% common and with hatch 300%)

https://github.com/GTNewHorizons/GTplusplus/assets/17906427/dbc8bd95-ee60-403a-8da1-e8a225668f25

### DIFF
--- a/src/main/java/gtPlusPlus/core/recipe/RECIPES_Machines.java
+++ b/src/main/java/gtPlusPlus/core/recipe/RECIPES_Machines.java
@@ -263,6 +263,7 @@ public class RECIPES_Machines {
         thermicFluidHeater();
         advHeatExchanger();
         chiselBuses();
+        solidifierHatches();
 
         gt4FarmManager();
         gt4Redstone();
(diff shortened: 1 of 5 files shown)


BlueWeabo:
What was the reason for giving it a 50% boost in speed?

Quetz4l:
I think it is logical that a mechanism with a specialized hatches works faster than a universal one

minecraft7771:
I think this would be better if it was a universal dual item/fluid hatch and placed in gregtech.

TheEpicGamer274:
I agree with mc here, a universal hatch/bus where you can select molds/shapes would be amazing for a bunch of multis and not just for this specific multi

BlueWeabo:
Yeah. I don't feel like a speed boost is warranted per say and having a dual item/fluid for any multi would be better imo.

TheEpicGamer274:
> I agree with mc here, a universal hatch/bus where you can select molds/shapes would be amazing for a bunch of multis and not just for this specific multi

Although the item extrusion is sorta covered by stocking busses, so mainly a fluid one for PAs, etc and not just LPF would be great

TheEpicGamer274:
> Yeah. I don't feel like a speed boost is warranted per say and having a dual item/fluid for any multi would be better imo.

agreed, lpf already has a speed bonus so whats the point in buffing it even more?

Quetz4l:
The problem is that multiblock recipe retrieval is separated by items and liquids. So I think it will be necessary to change the recipe search logic in all multiblocks. And it would be nice to make an api and have all multiblocks search through the api, not as it is now, where everyone has their own logic

BlueWeabo:
That is being worked on my @minecraft7771 and @miozune in base gt5u, which is why @minecraft7771 recommended putting a mixed fluid/item hatch/bus in gt5u instead of an addon mod. to make it easier to integrate with that.

chochem:
a generic mixed fluid+solid hatch/bus sounds a bit too easy and powerful imo. Like, who is gonna use the normal ones? rather have it here just for LPF.

bonus is not important, should be just fine without.

chochem:
unless the generic GT one has a strong restriction like items not being consumable, I would be against that.

edit: higher tiering can work of course

Alastors:
> unless the generic GT one has a strong restriction like items not being consumable, I would be against that.

I enjoy the concept myself, it introduces its own mix of solutions and issues to the player that make it just as interesting and arguably balanced as the non mixed hatches, and we could solve it being partially better than the mono hatches by just cutting how much it stores by 4 compared to a normal hatch of tier

Pxx500:
> > unless the generic GT one has a strong restriction like items not being consumable, I would be against that.
> 
> I enjoy the concept myself, it introduces its own mix of solutions and issues to the player that make it just as interesting and arguably balanced as the non mixed hatches, and we could solve it being partially better than the mono hatches by just cutting how much it stores by 4 compared to a normal hatch of tier

Reducing capacity of tanks is not really a nerf that late into the game since everything is automated, it's a penalty when a player does batch crafting by hand and has to re-insert the fluid multiple times

PR #678, opened 2023-06-24
Hydrogen loop fix

TheEpicGamer274:
should fix ticket https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/13786


miozune (approved):

Pilad:
Does this mean there is a mistake?
изображение

https://github.com/GTNewHorizons/GTplusplus/blob/2620d126f9cc24c99ea7545410d6c363e007ba00/src/main/java/gtPlusPlus/xmod/gregtech/loaders/recipe/RecipeLoader_ChemicalSkips.java#L104

chochem:
> Does this mean there is a mistake? изображение
> 
> https://github.com/GTNewHorizons/GTplusplus/blob/2620d126f9cc24c99ea7545410d6c363e007ba00/src/main/java/gtPlusPlus/xmod/gregtech/loaders/recipe/RecipeLoader_ChemicalSkips.java#L104

yes. very much so.

PR #680, opened 2023-06-24
Added LPF mode to WAILA tooltip

minecraft7771:


Steelux8 (approved):
More WAILA information is always good.

PR #683, opened 2023-06-29
Remove Breaker Box

miozune:
This thing has highly suspecious implementation. Since it has not been updated since 2019 nor has recipe, I think we can just remove it.
Also cleaned up `Meta_GT_Proxy`.


mitchej123 (approved):

Pilad:
> This thing has highly suspecious implementation. Since it has not been updated since 2019 nor has recipe, I think we can just remove it. Also cleaned up `Meta_GT_Proxy`.

Maybe it's worth removing blocks with the wrong texture from NEI? There are cases without textures and some items.

PR #693, opened 2023-07-17
Add QFT Water Line Skips

Steelux8:
This PR adds two QFT recipes related to the purified water line that is used to craft Optical Circuits. The first recipe is for a tier 2 QFT and crafts the first 4 grades of purified water, while the second recipe is for a tier 4 QFT and crafts the last 4 grades.

The normal recipes for this line are long (usually 250s), so I wasn't sure about how much water output to put on the 20s QFT recipes. I decided to distribute the outputs as 20kL of the first grade, 10kL of the second, 5kL of the third and 1kL of the fourth, for both recipes. Given how expensive the QFT is compared to the GT++ multis/PAs, this might be too slow, so I welcome feedback.

Tested and working on my GT++ dev env.


TheEpicGamer274 (approved):
I like the idea of having all of the tiers of water instead of being a direct use, still gives a use to the line if you want extra.

Runakai1 (approved):
Nice addition to the QFT. While not particularly a difficult line to get massive quantities for a Stargate without QFT, it's still a nice addition.

Dream-Master (approved):

PR #695, opened 2023-07-18
Particle Acceleration in the QFT

Steelux8:
This PR adds two new QFT recipes, related to complaints I've heard about Cyclotrons being a lag source in their single task of generating Unknown Particles, even with the DTPF recipe being a 1:1 conversion to Quantum Anomalies. The new tier 3 QFT recipe does it more quickly, and also generates a lot of Gravitons, due to a request by @GDCloudstrike to have a better source for those, due to godforge recipe plans.

<details>
  <summary>First recipe image</summary>

image
</details>

The production speed is not very good for UXV+ purposes, however. To remedy that, I decided to try to replace Quantum Anomalies in those super endgame tiers with something more advanced, starting with the new t4 QFT recipe that generates the super rare Cyclotron particles, like the Omega particles and the Higgs-Boson. From those, in the future, I'd like to add advanced, much more expensive anomalies that replace the low tier ones at that stage, and prevent their spam.

<details>
  <summary>Second recipe image</summary>

image
</details>

Finally, I buffed the Hydrogen Plasma Cyclotron recipe to generate a lot more Gravitons, the Beryllium Plasma recipe to have a much better chance of generating the super rare particles (would take over a year of constant run time on a single UV Cyclotron), and also added a copy of this recipe, with Celestial Tungsten Plasma, that lowers this time to, on average, 9-10 hours per Higgs-Boson particle.


Dream-Master:
GDCloud — 07/22/2023 7:35 PM
@Steelux | Chemistry & Powergen about the qft particle PR, I see you used rhenium and molybdenum ions in the higher tier catalyst recipe, but those currently aren't obtainable because their plasmas are uncraftable, so I see 3 options:
-change the ions in the recipe to other, craftable ones
-add recipes to rhenium and molybdenum plasma
-intentionally leave it as is and functionally lock them behind godforge
what do you think?

another issue misi pointed out was that beryllium plasma for the higher tier exotic particle recipe isnt craftable either, so the same questions apply

Steelux8:
I changed recipes so that the 3 plasmas that were necessary, but uncraftable, were replaced by craftable ones.

src/main/java/gtPlusPlus/xmod/gregtech/loaders/recipe/RecipeLoader_ChemicalSkips.java
```diff
@@ -260,10 +276,26 @@ private static void quantumTransformerRecipes() {
...
                 20 * 20,
                 (int) TierEU.RECIPE_UIV,
                 4);
+
+        // Rare Particles
+        CORE.RA.addQuantumTransformerRecipe(
+                new ItemStack[] { ItemUtils.getSimpleStack(GenericChem.mSynchrotronCapableCatalyst, 0),
+                        GregtechItemList.Laser_Lens_Special.get(1) },
+                new FluidStack[] { FluidUtils.getFluidStack("plasma.hydrogen", 30000),
+                        Materials.Helium.getPlasma(30000L), Materials.Beryllium.getPlasma(30000L),
```
GDCloudstrike: please change the beryllium plasma here too

GDCloudstrike (changes requested):

src/main/java/gtPlusPlus/xmod/gregtech/loaders/recipe/RecipeLoader_ChemicalSkips.java
Steelux8: Good catch, I forgot I had added the same plasma to the QFT recipe. Fixed.

GDCloudstrike (approved):

TheEpicGamer274 (commented):
Some of these recipe times seem a bit long, why?

Steelux8:
The recipes are long because the objective for new particles is the same as the original quantum anomaly: to obtain one of them, specifically the higgs-boson, and move to a better crafting option in the QFT.

meepmopper:
Depending on how these particles are implemented, such a long recipe time is likely to lead to qfts being spammed as they are insanely cheap lategame and it is much harder to generate more power to feed the 2/4 than to just place more of them. Maybe if the qft could achieve 4/4 oc by inputting a special fluid and had uncapped catalyst limit; this is just a suggestion for an idea. the qft implementing the neutron accelerator could be interesting as well.

Steelux8:
They are not supposed to be spammed. If such uses appear, I will revert them myself with the proper justification. The amount of quantum anomalies needed is not something that should be copied by these rarer particles.

PR #699, opened 2023-07-22
Fix EBF with Advanced Muffer Hatch being able to run without filter

Coolbeer:
Without a filter Advanced Muffler Hatch was unable to vent pollution because
damageAirFilter returns false if no filter is found.

So moved the call inside since the EBF is supposed to be able to run without a filter,
this way it will damage an inserted filter, but just continue if no filter inserted.

The pollution reduction is calculated elsewhere so this change shouldn't impact that.

### DIFF
--- a/src/main/java/gtPlusPlus/xmod/gregtech/api/metatileentity/implementations/GT_MetaTileEntity_Hatch_Muffler_Adv.java
+++ b/src/main/java/gtPlusPlus/xmod/gregtech/api/metatileentity/implementations/GT_MetaTileEntity_Hatch_Muffler_Adv.java
@@ -110,7 +110,8 @@ public class GT_MetaTileEntity_Hatch_Muffler_Adv extends GT_MetaTileEntity_Hatch
 
     @Override
     public boolean polluteEnvironment(MetaTileEntity unused) {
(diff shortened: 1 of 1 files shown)


draknyte1 (commented):
You realise this was probably intentional?

Dream-Master:
@Coolbeer so its not a fix its a buff. @draknyte1 can you explain why it was intended like this ?

Coolbeer:
While building the thing I was going by [this](https://gtnh.miraheze.org/wiki/Pollution), so therefore I expected it to work without a filter, there's also the pollution calculation in the code which checks for filter or not.

I do realize that the wiki might not be accurate, and that it's possible it's been changed to make the Advanced Muffler not work without a filter, but I don't really see a good reason for it not too, if it's deemed too OP in the current form maybe have it use the same data as a normal muffler block if run without a filter, I imagine the filter is basically slotted in somewhere and removing it just makes the pollution go straight out the muffler instead of through the filter.

chochem:
this is meant to use filters, as alk pointed out. If you don't want a filter you can just use the normal one, so I think that makes a lot of sense how it currently works.

Dream-Master:
More people vote against it

PR #706, opened 2023-07-25
Fixed mass fab sometimes crashing in scrap mode

minecraft7771:
Also fixes it trying to consume ghost circuits, which results in no output and the machine being deadlocked.


src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntity_MassFabricator.java
```diff
@@ -230,20 +230,27 @@ public CheckRecipeResult process() {
...
-                            40,
-                            MaterialUtils.getVoltageForTier(1),
-                            0);
-                    return FindRecipeResult.ofSuccess(recipe);
+                    if (inputItems != null) {
+                        for (ItemStack item : inputItems) {
+                            if (item == null || item.stackSize == 0) continue;
+                            ItemStack aPotentialOutput = GT_ModHandler
+                                    .getRecyclerOutput(GT_Utility.copyAmount(1, item), 0);
+                            if (aPotentialOutput == null) continue;
```
miozune: Isn't it supposed to consume input and output nothing, instead of just ignoring it?

miozune (approved):

PR #709, opened 2023-07-29
Make sure to account for first output being chanced on steam grinder

BlueWeabo:


src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/processing/steam/GregtechMetaTileEntity_SteamMacerator.java
```diff
@@ -138,11 +138,10 @@ protected ProcessingLogic createProcessingLogic() {
             public GT_ParallelHelper createParallelHelper(@Nonnull GT_Recipe recipe) {
                 return super.createParallelHelper(recipe).setCustomItemOutputCalculation(parallel -> {
                     ArrayList<ItemStack> items = new ArrayList<>();
-                    ItemStack output = recipe.getOutput(0);
-                    if (output != null) {
-                        output = output.copy();
-                        output.stackSize *= parallel;
-                        items.add(output);
+                    for (int i = 0; i < parallel; i++) {
+                        if (recipe.getOutputChance(0) > getBaseMetaTileEntity().getRandomNumber(10000)) {
```
miozune: `GT_ParallelHelper` uses `XSTR`, so using it here too would be better.
Also `GT_ParallelHelper` should be fixed to use `>` instead of `>=`; Currently output chances are always 0.01 point higher

miozune (approved):

PR #713, opened 2023-08-03
Lower reservoir hatch tier to EV, add tooltip showing tier for reservoir + air intake hatches

querns:
The reservoir hatch currently has an internal tier of LuV. This means that it can't be used in a chemplant that has lower than LuV hulls. Since the reservoir hatch is made in EV, I lowered its tier to EV. 

java_Smwoj8ri5u

I also added to the tooltip in a base class to show the tier of the hatch. Affects: Reservoir Hatch, Air Intake Hatch, Extreme Air Intake Hatch.

image

image

### DIFF
--- a/src/main/java/gtPlusPlus/xmod/gregtech/api/metatileentity/implementations/GT_MetaTileEntity_Hatch_FluidGenerator.java
+++ b/src/main/java/gtPlusPlus/xmod/gregtech/api/metatileentity/implementations/GT_MetaTileEntity_Hatch_FluidGenerator.java
@@ -3,11 +3,13 @@ package gtPlusPlus.xmod.gregtech.api.metatileentity.implementations;
 import net.minecraft.entity.player.EntityPlayer;
 import net.minecraft.item.ItemStack;
 import net.minecraft.nbt.NBTTagCompound;
+import net.minecraft.util.EnumChatFormatting;
 import net.minecraft.world.World;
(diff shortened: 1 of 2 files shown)


src/main/java/gtPlusPlus/xmod/gregtech/api/metatileentity/implementations/GT_MetaTileEntity_Hatch_FluidGenerator.java
```diff
@@ -46,10 +48,14 @@ public GT_MetaTileEntity_Hatch_FluidGenerator(final String aName, final int aTie
     @Override
     public synchronized String[] getDescription() {
         mDescriptionArray[1] = "Capacity: " + GT_Utility.formatNumbers(getCapacity()) + "L";
+        final String[] hatchTierString = new String[] {
+                "Hatch Tier: " + GT_Values.TIER_COLORS[mTier] + GT_Values.VN[mTier] + EnumChatFormatting.RESET };
```
minecraft7771: Use `GT_Utility.getColoredTierNameFromTier`

querns: Done.

minecraft7771 (approved):

PR #716, opened 2023-08-08
More LFTR Chain Fixes

Steelux8:
Coming in with more fixes for the LFTR chain before the next stable, including some problems that resurfaced from processing logic changes:

- LFTR no longer keeps running without fuel;
- Sparge Tower no longer runs 10x as fast as NEI's recipe times, NEI was adjusted to reflect the real processing times;
- Sparge Tower now consumes EU properly, instead of "generating" it;
- Fuel Refinery now has more reasonable hatch requirements, while keeping the old maximums for old structures;
- Tweaked the very long (7200s-9000s) Refinery recipes to only do 1/10 of the recipe, but at 1/10 of the recipe time, for more reasonable fuel refining.

### DIFF
--- a/src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/processing/GregtechMetaTileEntity_SpargeTower.java
+++ b/src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/processing/GregtechMetaTileEntity_SpargeTower.java
@@ -224,6 +224,11 @@ public class GregtechMetaTileEntity_SpargeTower extends GregtechMeta_MultiBlockB
                             tRecipe.mFluidInputs[1]);
                     this.mOutputFluids = aFluidOutputs.toArray(new FluidStack[0]);
                     updateSlots();
+
+                    if (lEUt > 0) {
+                        lEUt = (-lEUt);
+                    }
+
                     return CheckRecipeResultRegistry.SUCCESSFUL;
                 }
             }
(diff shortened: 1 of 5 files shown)


src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMTE_NuclearReactor.java
```diff
@@ -359,6 +362,91 @@ protected GT_OverclockCalculator createOverclockCalculator(@NotNull GT_Recipe re
...
+                    if (!result.wasSuccessful()) {
+                        return result;
+                    } else {
+                        lastRecipe = recipe;
+                    }
+                } else {
+                    if (findRecipeResult.getState() == FindRecipeResult.State.INSUFFICIENT_VOLTAGE) {
+                        return CheckRecipeResultRegistry.insufficientPower(findRecipeResult.getRecipeNonNull().mEUt);
+                    } else {
+                        resetMultiProcessing();
```
miozune: How about this instead of just copying them all?
```java
CheckRecipeResult result = super.process();
if (!result.wasSuccessful()) {
    resetMultiProcessing();
}
return result;
```

src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/processing/GregtechMetaTileEntity_SpargeTower.java
```diff
@@ -224,6 +224,11 @@ public boolean isCorrectMachinePart(ItemStack aStack) {
                             tRecipe.mFluidInputs[1]);
                     this.mOutputFluids = aFluidOutputs.toArray(new FluidStack[0]);
                     updateSlots();
+
+                    if (lEUt > 0) {
```
miozune, line 229: Since this value is stored in existing machines, modify `loadNBTData` to also make it negative with comment

src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMTE_NuclearReactor.java
Steelux8: Good shout, I should have called super in this way.

miozune (approved):

PR #724, opened 2023-08-20
Remove batch mode from steam multis

miozune:
It was never a thing


BlueWeabo (approved):
not that it would work now. (unless somehow a recipe was under 128 ticks. prolly some out there)

miozune:
I mean this also disables batch mode button on GUI

BlueWeabo:
Yeah makes sense. Not needed if it does nothing.

PR #725, opened 2023-08-21
Enable batch mode on QFT

S4mpsa:
Because there is no reason it should not be enabled when you can relatively easy 1-tick it with laser hatches.


TheEpicGamer274 (approved):

BlueWeabo (approved):
Batching quantum time itself

PR #749, opened 2023-09-18
Add Voltage-CoilHeat adjustment for Volcanus

RealSilverMoon:
As is well known, EBF has the feature of adjusting heat based on voltage, but Volcanus does not. This feature is generally not very useful, except here: UHV Superconductor. You need UHV Superconductor to make UHV Energy Hatch, which are required for crafting T8 rockets, obtaining the Awakened Dragon, and making Awakened Draconium Coils. The problem is that UHV Superconductors require Awakened Draconium Ingots to smelt. Therefore, you have to use a EBF (or you have a Electrum Flux Coil Mega EBF) to bypass this peculiar situation. Adding heat adjustment to Volcanus can avoid the need for this EBF, which is only needed to smelt a few UHV superconductor ingots then got dismantled.
image

### DIFF
--- a/src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/processing/advanced/GregtechMetaTileEntity_Adv_EBF.java
+++ b/src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/processing/advanced/GregtechMetaTileEntity_Adv_EBF.java
@@ -61,7 +61,8 @@ public class GregtechMetaTileEntity_Adv_EBF extends GregtechMeta_MultiBlockBase<
     private int mCasing;
     private final ArrayList<GT_MetaTileEntity_Hatch_CustomFluidBase> mPyrotheumHatches = new ArrayList<>();
 
-    private HeatingCoilLevel mHeatingCapacity = HeatingCoilLevel.None;
+    private HeatingCoilLevel mHeatingLevel = HeatingCoilLevel.None;
+    private int mHeatingCapacity=0;
 
     public GregtechMetaTileEntity_Adv_EBF(int aID, String aName, String aNameRegional) {
         super(aID, aName, aNameRegional);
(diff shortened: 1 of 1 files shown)


Pxx500:
Seems like I misunderstood the PR intentions, but still a balance change. Though I don't think it's major now

chochem:
will you also change the documentation?

chochem:
One thing the PR _curiously_ fails to mention is that this also gives an additional heat OC in many situations. This is certainly a balance change, a clear buff. I don't care that much as long as they fix documentation accordingly. But Volcanus certainly isn't in a dire need for buffs.

RealSilverMoon:
> One thing the PR _curiously_ fails to mention is that this also gives an additional heat OC in many situations. This is certainly a balance change, a clear buff. I don't care that much as long as they fix documentation accordingly. But Volcanus certainly isn't in a dire need for buffs.

This is because I didn't send the voltage extra heat into the createOverclockCalculator, so Volcanus doesn't actually gain extra OC from this heat. It is certain that this may seem different from the EBF, but we can note this in the documentation (voltage heat doesn't provide OC). Anyway, even with UV, not so much recipes can get an extra OC from the additional 800K heat, and beyond UHV it's mega EBF and DTPF time.

chochem (changes requested):
thank you for clarifying!

However I have to say that makes this a lot worse. You take the perfectly fine system of one heat value and split it into 2 heat values, only one of which is shown to the player. Everybody will understandably just assume its bugged. I will have to demand changes here, this is too messy.

In my opinion we don't need these changes at all but I am happy to wait for other devs to chime in on that front. But the current state is not ok for sure.

Steelux8:
If this recipe is the problem, then I suggest changing its heat value so that the Volcanus can do it with Electrum Flux coils, instead of changing the multi's mechanics to accomodate this.

chochem:
alright. I had a chat with a handful of devs. There is a clear consensus that we don't want the Volcanus changed as it is in a good spot as is. So I will close this PR.

Even if we were ever going to change something here it would at best be the recipe, not the multi. As steelux says. But even that might remain, the opinions on that are more varied. Basically, you should consider making an MBF at that stage of course but can also use a EBF if you prefer.

This is not meant to discourage you from contributing. I guess you did not put in too much work here, but it is always good advice to have some discussion about changes beforehand if they are more than a bug fix. Otherwise it can sometimes happen that the change isn't wanted.

PR #750, opened 2023-09-19
Fix incorrect extruder recipe voltages for GT++ materials.

fluffle:
Fixes a few recipe problems I ran into while adding pipe casing recipes. This should be fine to merge even if the planned EHE changes end up taking a different path.

### DIFF
--- a/src/main/java/gtPlusPlus/core/recipe/RECIPES_General.java
+++ b/src/main/java/gtPlusPlus/core/recipe/RECIPES_General.java
@@ -25,6 +25,7 @@ import gtPlusPlus.core.item.ModItems;
 import gtPlusPlus.core.lib.CORE;
 import gtPlusPlus.core.material.ALLOY;
 import gtPlusPlus.core.material.ELEMENT;
+import gtPlusPlus.core.material.Material;
 import gtPlusPlus.core.recipe.common.CI;
 import gtPlusPlus.core.util.minecraft.FluidUtils;
 import gtPlusPlus.core.util.minecraft.ItemUtils;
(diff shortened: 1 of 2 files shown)


Dream-Master:
@fluffle are you remove the void metal pipes ?

fluffle:
I shouldn't be, no? I haven't tested with *just* this code explicitly, but the follow-on PR that added recipes for the pipe casing blocks still had void pipes when I tested it.

The problem is that the `GT_Materials` enum doesn't store the processing voltage. To correct the pipe extruder recipes always being either MV or HV, I changed all the GT alloys to create their respective GT pipes using data from the GT++ `Material` class instead. But void isn't a GT++ thing, so it doesn't *have* a GT++ `Material`. To get around this I pulled the thaumcraft check out of the loop and created the void pipes manually using the `GT_Materials` enum value.

Dream-Master:
Ok will check you pr later

chochem:
GT materials also have processing voltages. why not use the proper gt stuff?

Dream-Master:
> GT materials also have processing voltages. why not use the proper gt stuff?

didnt we do stuff like that in gt as well? to autogenerate recipes? But idk what are the math exactly make with the process time and tier

miozune (approved):

fluffle:
> GT materials also have processing voltages. why not use the proper gt stuff?

I don't think these *are* "proper" GT materials. I could be missing something, since i'm not super familiar with the codebase, but these materials are defined in a sub-enum of the GregtechOrePrefixes enum here:

https://github.com/GTNewHorizons/GTplusplus/blob/0209cf0c2610cdac0a8a379454340c0481f36466/src/main/java/gtPlusPlus/xmod/gregtech/api/enums/GregtechOrePrefixes.java#L372

This enum only includes a small subset of the materials GT defines and a subset of the alloys GT++ defines, and looks like it is there to bridge these two worlds... sort of. Frankly its existence confuses TF out of me but old codebases with many maintainers over time regularly have these sort of warts ;-)

PR #751, opened 2023-09-19
Seaweed QFT recipe

GDCloudstrike:
Adds a TCeti seaweed line skip to the QFT which outputs seaweed extract and iodine (a seaweed related element) directly, skipping the processing chain unlocked in UHV. 

This change aims to retire the biovat from endgame completely, as seaweed was the last thing requiring it. The QFT recipe requires a t4 QFT meaning it is unlocked in UMV, same as the bio line skip.

Catalyst recipe:
image

QFT recipe:
image


TheEpicGamer274 (approved):
Seems fair to me

PR #752, opened 2023-09-20
Some more RA2 conversions

chochem:
Some more RA2 conversions with bits of code clean-up. Also updated for the deprecations of noInputFluid() and similar.


Dream-Master (approved):
Checked the recipes by reading it here. Seems fine to me.

PR #753, opened 2023-09-20
Change Alloy Blast Smelter recipes per page to 2

querns:
While I wasn't able to quite stick the landing on sorting the recipes for ABS alloys such that the proper recipe for making alloys was the first one seen, increasing the recipes per page to 2 should allow people who don't have their UI scale set to "Geriatric" the ability to see the proper recipe on the first page anyway.


Pxx500 (approved):
looks fine, tested in-game

PR #754, opened 2023-09-22
Add Alternative Recipe in LFTR Chain to Avoid Cells

Steelux8:
<details>
  <summary>Screenshots</summary>
  

  
</details>

See above for the screenshots of the relevant recipes.

The LFTR chain has an EBF recipe that handles multiple fluid inputs and outputs. However, as the EBF cannot normally handle so many fluids, most of them are in cells, which should not be present in IV+ automation. For that reason, I added an equivalent Chemical Plant recipe that uses no cells.

### DIFF
--- a/src/main/java/gtPlusPlus/xmod/gregtech/loaders/recipe/RecipeLoader_NuclearFuelProcessing.java
+++ b/src/main/java/gtPlusPlus/xmod/gregtech/loaders/recipe/RecipeLoader_NuclearFuelProcessing.java
@@ -213,6 +213,22 @@ public class RecipeLoader_NuclearFuelProcessing {
                 MaterialUtils.getVoltageForTier(5),
                 5400);
 
+        CORE.RA.addChemicalPlantRecipe(
+                new ItemStack[] { },
+                new FluidStack[] {
+                        FLUORIDES.URANIUM_HEXAFLUORIDE.getFluidStack(1000),
+                        NUCLIDE.LiFBeF2.getFluidStack(1000),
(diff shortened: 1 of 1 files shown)


Dream-Master:
To be honest i would remove the ebf recipes because it make not much sense (or deprecate it) some day we will define the checmical plant as a own machine with more unique recipes and a need for progress. Maybe let it in for now but we need work on this some day.

BlueWeabo:
Technically you could just make the ebf recipe have two fluid inputs and two fluid outputs and it won't break :trollface:

Dream-Master:
But cooking fluids in ebf its a bit wired. its more a chemical recipe in LCR or chem plant and not in the ebf.
If this fluid is some molten stuff like a metal this would be a different thing. @BlueWeabo

Dream-Master (approved):
lets add it for now.

Steelux8:
> To be honest i would remove the ebf recipes because it make not much sense (or deprecate it) some day we will define the checmical plant as a own machine with more unique recipes and a need for progress. Maybe let it in for now but we need work on this some day.

I only kept the EBF recipe because removing it is a breaking change to existing setups.

Dream-Master:
> > To be honest i would remove the ebf recipes because it make not much sense (or deprecate it) some day we will define the checmical plant as a own machine with more unique recipes and a need for progress. Maybe let it in for now but we need work on this some day.
> 
> I only kept the EBF recipe because removing it is a breaking change to existing setups.

ik and thats why i merge it before.

PR #760, opened 2023-10-10
Make soldering iron change multiblock void mode

LewisSaber:
image

### DIFF
--- a/src/main/java/gtPlusPlus/xmod/gregtech/api/metatileentity/implementations/base/GregtechMeta_MultiBlockBase.java
+++ b/src/main/java/gtPlusPlus/xmod/gregtech/api/metatileentity/implementations/base/GregtechMeta_MultiBlockBase.java
@@ -1174,7 +1174,6 @@ public abstract class GregtechMeta_MultiBlockBase<T extends GT_MetaTileEntity_Ex
     @Override
     public boolean onSolderingToolRightClick(ForgeDirection side, ForgeDirection wrenchingSide, EntityPlayer aPlayer,
             float aX, float aY, float aZ) {
-        super.onSolderingToolRightClick(side, wrenchingSide, aPlayer, aX, aY, aZ);
         if (supportsVoidProtection()) {
             Set<VoidingMode> allowed = getAllowedVoidingModes();
             setVoidingMode(getVoidingMode().nextInCollection(allowed));
@@ -1183,8 +1182,7 @@ public abstract class GregtechMeta_MultiBlockBase<T extends GT_MetaTileEntity_Ex
                     StatCollector.translateToLocal("GT5U.gui.button.voiding_mode") + " "
                             + StatCollector.translateToLocal(getVoidingMode().getTransKey()));
             return true;
-        }
-        return false;
+        } else return super.onSolderingToolRightClick(side, wrenchingSide, aPlayer, aX, aY, aZ);
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


src/main/java/gtPlusPlus/xmod/gregtech/api/metatileentity/implementations/base/GregtechMeta_MultiBlockBase.java
```diff
@@ -1170,6 +1171,20 @@ public boolean onWireCutterRightClick(ForgeDirection side, ForgeDirection wrench
...
+            float aX, float aY, float aZ) {
+        if (supportsVoidProtection()) {
+            Set<VoidingMode> allowed = getAllowedVoidingModes();
+            setVoidingMode(getVoidingMode().nextInCollection(allowed));
+            GT_Utility.sendChatToPlayer(
+                    aPlayer,
+                    StatCollector.translateToLocal("GT5U.gui.button.voiding_mode") + " "
+                            + StatCollector.translateToLocal(getVoidingMode().getTransKey()));
+        }
+        return true;
```
Glease: return true only if you did something

src/main/java/gtPlusPlus/xmod/gregtech/api/metatileentity/implementations/base/GregtechMeta_MultiBlockBase.java
```diff
@@ -1170,6 +1171,20 @@ public boolean onWireCutterRightClick(ForgeDirection side, ForgeDirection wrench
         return false;
     }
 
+    @Override
+    public boolean onSolderingToolRightClick(ForgeDirection side, ForgeDirection wrenchingSide, EntityPlayer aPlayer,
+            float aX, float aY, float aZ) {
+        if (supportsVoidProtection()) {
```
Glease, line 1177: call superclass `onSolderingToolRightClick`, even if it do nothing. it will save a lot of headache if we ever add any features to superclass involving solder right click

Glease (changes requested):

src/main/java/gtPlusPlus/xmod/gregtech/api/metatileentity/implementations/base/GregtechMeta_MultiBlockBase.java
```diff
@@ -1174,15 +1174,17 @@ public boolean onWireCutterRightClick(ForgeDirection side, ForgeDirection wrench
     @Override
     public boolean onSolderingToolRightClick(ForgeDirection side, ForgeDirection wrenchingSide, EntityPlayer aPlayer,
             float aX, float aY, float aZ) {
+        super.onSolderingToolRightClick(side, wrenchingSide, aPlayer, aX, aY, aZ);
```
Glease:
```suggestion
if (super.onSolderingToolRightClick(side, wrenchingSide, aPlayer, aX, aY, aZ)) {
    return
}
``` 

This means if superclass did anything, we will bail out early instead of letting one user action translate into two change in the controller

Glease (changes requested):

src/main/java/gtPlusPlus/xmod/gregtech/api/metatileentity/implementations/base/GregtechMeta_MultiBlockBase.java
minecraft7771: Wouldn't it make more sense to only call super if we did nothing here. Because otherwise there could in theory be the situation that we can never do our thing because super always soles something. That would be against the idea of why we do overrides.

LewisSaber:
I think minecraft7771 is right here

Glease (approved):
i guess we will just deal with gtpp by then

PR #761, opened 2023-10-11
Clean up BOP_block_Registrator and convert it to RA2

chochem:
Just a small cleanup with RA2 conversion.

- converted recipes to RA2
- removed duplicate sapling recipes (all are added by GT too)
- removed duplicate Pulverisation and Lathe recipes (all are added by GT too)
- removed unused themal expansion sawmill compat
- use proper GT method for Pyrolyse Oven recipes (instead of an outdated copied over variant with wrong circuits)
- some minor code cleanup

### DIFF
--- a/src/main/java/gtPlusPlus/xmod/bop/blocks/BOP_Block_Registrator.java
+++ b/src/main/java/gtPlusPlus/xmod/bop/blocks/BOP_Block_Registrator.java
@@ -120,12 +120,6 @@ public class BOP_Block_Registrator {
                     final ItemStack tPlanks = GT_Utility.copy(tStack);
                     tPlanks.stackSize = tPlanks.stackSize * 3 / 2;
                     GT_Values.RA.stdBuilder().itemInputs(new ItemStack(aStack.getItem(), 1, i))
-                            .fluidInputs(Materials.Lubricant.getFluid(1L))
-                            .itemOutputs(
-                                    GT_Utility.copy(tPlanks),
-                                    GT_OreDictUnificator.get(OrePrefixes.dust, Materials.Wood, 1L))
-                            .duration(10 * SECONDS).eut(8).addTo(sCutterRecipes);
-                    GT_Values.RA.stdBuilder().itemInputs(new ItemStack(aStack.getItem(), 1, i))
                             .itemOutputs(
                                     GT_Utility.copyAmount(
                                             GT_Mod.gregtechproxy.mNerfedWoodPlank ? ((long) tStack.stackSize)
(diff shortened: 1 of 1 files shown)


src/main/java/gtPlusPlus/xmod/bop/blocks/BOP_Block_Registrator.java
```diff
@@ -136,31 +119,25 @@ public static final void addLogRecipes(final ItemStack aStack) {
...
+                                    GT_OreDictUnificator.get(OrePrefixes.dust, Materials.Wood, 1L))
+                            .duration(10 * SECONDS).eut(8).addTo(sCutterRecipes);
+                    GT_Values.RA.stdBuilder().itemInputs(new ItemStack(aStack.getItem(), 1, i))
+                            .itemOutputs(
+                                    GT_Utility.copyAmount(
+                                            GT_Mod.gregtechproxy.mNerfedWoodPlank ? ((long) tStack.stackSize)
+                                                    : (((long) tStack.stackSize) * 5L / 4),
+                                            tStack),
+                                    GT_OreDictUnificator.get(OrePrefixes.dust, Materials.Wood, 2L))
+                            .duration(10 * SECONDS).eut(8).addTo(sCutterRecipes);
```
BlueWeabo, line 129: Should this have no fluid input? Wasn't this water? Because otherwise the machines could sometimes just do this recipes instead of the one with lubricant

src/main/java/gtPlusPlus/xmod/bop/blocks/BOP_Block_Registrator.java
```diff
@@ -171,44 +148,29 @@ public static final void addLogRecipes(final ItemStack aStack) {
...
+                                GT_OreDictUnificator.get(OrePrefixes.dust, Materials.Wood, 1L))
+                        .duration(10 * SECONDS).eut(8).addTo(sCutterRecipes);
+                GT_Values.RA.stdBuilder().itemInputs(GT_Utility.copyAmount(1L, aStack))
+                        .itemOutputs(
+                                GT_Utility.copyAmount(
+                                        GT_Mod.gregtechproxy.mNerfedWoodPlank ? ((long) tStack2.stackSize)
+                                                : (((long) tStack2.stackSize) * 5L / 4),
+                                        tStack2),
+                                GT_OreDictUnificator.get(OrePrefixes.dust, Materials.Wood, 2L))
+                        .duration(10 * SECONDS).eut(8).addTo(sCutterRecipes);
```
BlueWeabo, line 156: Same as above. Shouldn't water be an input?

src/main/java/gtPlusPlus/xmod/bop/blocks/BOP_Block_Registrator.java
chochem: I just ra2ed that part :P
no fluid should generate the 3 water/distilled/lubricant recipes (its the rare case where RA2 is not explicit).
but that begs the question what the lubricant one is for. will remove that one

BlueWeabo (approved):
Thanks for the cleanup

PR #767, opened 2023-10-17
Cleaning up

Connor-Colenso:


boubou19 (approved):
let's find out what breaks :kekw:

chochem:
why would you do that boubou

boubou19:
a 23k lines PR no one will correctly review will sit here basically forever until someone decides to merge it. Given the size of the PR, merge conflicts would happen soon if it was still open. + Colen used it in full pack and it did not break.

chochem:
forever is anything more than 5 minutes?

anyway here are some obvious problems:
- why is round robinator removed? discord search says people are using it.
- why are ULV gens removed? can run simple washer with that. (and people are)
- why are the chunkloaders removed?
- why are the catalyst methods removed? they look useful.

Pilad:
Remove from NEI gt++ workbench
https://github.com/GTNewHorizons/GTplusplus/blob/fbdbb388c369fae549cfde3eed38b536f7461d90/src/main/java/gtPlusPlus/nei/NEI_IMC_Sender.java#L80
80-81-82
изображение

PR #770, opened 2023-10-21
asynchronous cape download

Glease:


boubou19:
didn't know gt++ had capes. What cape is used in case the player has both a gt++ cape and nh cape?

Glease:
presumably both, since I don't see either of the two cape renderers checking for each other
or maybe alk simply won't add one to gt++ cape list if it's on gt cape list

OneEyeMaker (approved):

draknyte1:
> presumably both, since I don't see either of the two cape renderers checking for each other
> or maybe alk simply won't add one to gt++ cape list if it's on gt cape list

I provided a config option as the simplest solution. Greg wasn’t adding new cape wearers and it’s impossible to look at other mods to determine if they provide capes. I don’t think any mod providing capes does more than this, if at all.

PR #772, opened 2023-10-21
add back classes referenced by code injected with asm

Glease:


Alexdoru (approved):

Alexdoru:
@GTNH-Colen be careful when you delete stuff, some code that shows as unused can in fact be called by code genereted with ASM during runtime.

Connor-Colenso:
Yea I wasn't aware that was possible before, good to know.

PR #776, opened 2023-11-03
Fix Calcium Chloride recipe to generate in LCR

### LINKED ISSUE #14768: Add LCR recipe for calcium cabonate dust from GT++

danzah

### Your Pack Version

2.4.0

### Your Proposal

Calcium cabonate dust from GT++  is a byproduct of processing in gt++ machines. It can be processed in a chemical reactor, but it cannot be done in LCR
изображение_2023-10-24_223423862

### Your Goal

To avoid using small chemical reactor

### Your Vision

1) Convertation calcium carbonate to calcite with crafting table(chemical formula is same)
2) chenge calcium carbonate id on calcite id (due to the reason described in paragraph 1)

### Final Checklist

- [x] I have searched this issue tracker and there is nothing similar already. Posting on a closed issue saying *I like this feature please reconsider adding it* will prompt us to investigate and reopen it once we confirm your report.
- [X] I believe there is nothing similar in the pack already, or the existing solution isn't good enough.
- [X] I understand this change request may not attract enough attention and thus not be implemented.
- [X] I understand this change request may be rejected due to other community members thinking it's inappropriate.
- [X] I believe this feature would make the pack better.

chochem:
changed c18 -> c8 for calcium carbonate to calcium chloride recipes.
It now generates the MB recipe as expected (I checked in full pack).

Fixes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/14768

### DIFF
--- a/src/main/java/gtPlusPlus/xmod/gregtech/loaders/recipe/RecipeLoader_GlueLine.java
+++ b/src/main/java/gtPlusPlus/xmod/gregtech/loaders/recipe/RecipeLoader_GlueLine.java
@@ -4,6 +4,8 @@ import static gregtech.api.enums.Mods.BartWorks;
 import static gregtech.api.enums.Mods.GalacticraftCore;
 import static gregtech.api.enums.Mods.NewHorizonsCoreMod;
 import static gregtech.api.enums.Mods.TecTech;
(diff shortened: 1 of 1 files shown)


Ethryan (approved):
Looks ok

PR #778, opened 2023-11-06
Remove WA leftovers

miozune:
Well we still don't remove WA from the game, these are leftover of copies from NHCore


Alastors (approved):
Looks good to me.

PR #785, opened 2023-11-27
Reduce minimum casing requirement for various machines

Pelotrio:
This pull request reduces the min casing amount of various machine to make it consistent with other multiblocks of equal size. As a baseline the multi with the least amount of hatches has been used. 

**3x3x3:**
LPF 6 -> 6 (unchanged and baseline)
Industrial Material Press 10 -> 6
Industrial Sledgehammer 10 -> 6
Industrial Electrolyzer 10 -> 6
Industrial Centrifuge 10 -> 6
Industrial Arc Furnace 10 -> 6

**3x3x5**
Industrial Extrusion machine 14 -> 14 (unchanged and baseline)
Wire Factory 20 -> 14
Cutting Factory 26 -> 14

### Why
With the current amount of different components needed its mostly just an annoyance to have a high requirement of casings as you are limited by how many different hatches you can use (especially relevant with cribs). The casing themselves are also not expensive either. Multis like the Industrial Material Press suffer greatly from that as it is used quite a lot. In my opinion it would make sense that multis that are equal in size (not counting any special blocks or coils) have the same min casing requirement.

### DIFF
--- a/src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/processing/GregtechMetaTileEntity_IndustrialArcFurnace.java
+++ b/src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/processing/GregtechMetaTileEntity_IndustrialArcFurnace.java
@@ -76,7 +76,7 @@ public class GregtechMetaTileEntity_IndustrialArcFurnace extends
                 .addInfo("Right-click controller with a Screwdriver to change modes")
                 .addInfo("Max Size required to process Plasma recipes").addPollutionAmount(getPollutionPerSecond(null))
                 .addSeparator().addController("Top center").addStructureInfo("Size: nxnx3 [WxHxL] (Hollow)")
-                .addStructureInfo("n can be 3, 5 or 7").addCasingInfoMin(mCasingName, 10, false)
+                .addStructureInfo("n can be 3, 5 or 7").addCasingInfoMin(mCasingName, 6, false)
                 .addInputBus("Any Casing", 1).addOutputBus("Any Casing", 1).addInputHatch("Any Casing", 1)
                 .addOutputHatch("Any Casing", 1).addEnergyHatch("Any Casing", 1).addMaintenanceHatch("Any Casing", 1)
                 .addMufflerHatch("Any Casing", 1).toolTipFinisher(CORE.GT_Tooltip_Builder.get());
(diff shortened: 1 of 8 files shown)


boubou19:
why?

Dream-Master:
I guess that we can use more hatches ?

Pelotrio:
> I guess that we can use more hatches ?

Yes that is the main reason and for consistency with other equal sized multis.

bombcar:
> I guess that we can use more hatches ?

There's also a cost issue if a multi needs casings of "incredibly complex hard to make" but will take infinite ULV hatches, then you can cheap out by going hatch-crazy.

Too bad we can't have cleanroom like mechanics where "you can have one LV hatch, two MV, and infinite of correct tier"

Pelotrio:
> > I guess that we can use more hatches ?
> 
> There's also a cost issue if a multi needs casings of "incredibly complex hard to make" but will take infinite ULV hatches, then you can cheap out by going hatch-crazy.
> 
> Too bad we can't have cleanroom like mechanics where "you can have one LV hatch, two MV, and infinite of correct tier"

That is partly true but the cost of all the casings of the multis I changed are very negligible. Even if this would only be relevant at the very beginning when you first unlock the multi and want to exploit this mechanic. This is also only a follow up to multiple prs that already reduced the min casing amount for many gtpp multis.

chochem:
the arc furnace is not 3x3x3. anyway I dont see any reason for this to be changed.

Pelotrio:
> the arc furnace is not 3x3x3. anyway I dont see any reason for this to be changed.

@chochem the arc furnace is 3x3x3 by default if you dont increase its size. The min casing amount for the different sizes was the same so I left it at that as well.

chochem:
you didnt leave it, no. its changed currently

Pelotrio:
Yes that is not what I meant. The arc furnace is 3x3x3 if you build the smallest size and has a min casing amount of 10 like a few other multiblocks of the same size. The min casing amount doesnt get increased if you build it larger and stays at 10. I changed all all of them to be six like the lpf because the current implementation does not change the min casing amount with size. It is still changed because the smallest size is 3x3x3 but I can also remove/change that if its not wanted.

tth05:
> the arc furnace is not 3x3x3. anyway I dont see any reason for this to be changed.

Can you please read the "Why" section and then provide a proper reason why you're against it?

Also, this change is similar to this https://github.com/GTNewHorizons/GTplusplus/pull/481 which was fine

Dream-Master (approved):

PR #796, opened 2023-12-09
Fix some multis incorrectly restricting fluid input for hatches

miozune:
Requires https://github.com/GTNewHorizons/GT5-Unofficial/pull/2400

### DIFF
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -1,5 +1,5 @@
 dependencies {
-    api('com.github.GTNewHorizons:GT5-Unofficial:5.09.44.105-pre:dev')
+    api('com.github.GTNewHorizons:GT5-Unofficial:5.09.44.106:dev')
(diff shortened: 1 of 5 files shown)


dependencies.gradle
```diff
@@ -1,5 +1,5 @@
 dependencies {
-    api('com.github.GTNewHorizons:GT5-Unofficial:5.09.44.104:dev')
+    api('com.github.GTNewHorizons:GT5-Unofficial:5.09.44.105-pre:dev')
```
Connor-Colenso: Aren't we supposed to not depend on -pre versions due to periodic purging?

Connor-Colenso (changes requested):

dependencies.gradle
Dream-Master: i fix that one sec @Connor-Colenso just spprove it i make a gt tag

Connor-Colenso (approved):

PR #799, opened 2023-12-12
Fix asymmetry in the Quantum Force Transformer

Pelotrio:
Before:
image

I cant get a good screenshot from after it because the structure preview is broken in dev but I checked in game and its correct.
image
image


LewisSaber (approved):

S4mpsa:
This breaks everyones existing QFTs. Should probably inform this somewhere so everyone past UIV aren't blindsided why their QFTs suddenly broke...

Dream-Master:
@Pelotrio just reopen the pr and we merge this after stable

PR #805, opened 2023-12-15
Fix QFT focus chance calculation

### LINKED ISSUE #15082: QFT focusing outputs calculated incorrectly

step5

### Your Pack Version

2.4.0

### Your Server

Private server

### Java Version

Java 20

### Type of Server

None

### Your Expectation

QFT outputs are boosted with Neptunium focusing according to the calculation mentioned in the quest book. The total chance should add up to 100%.

image

For example, for a 3-output recipe and same-tier focusing blocks, the chances should be:
33%/33%/33% -> **66%**/16%/16%

### The Reality

The chance for the boosted output is calculated incorrectly, which makes it always less than what it should be. 
From the [codes](https://github.com/GTNewHorizons/GTplusplus/blob/7a25aa1a3324af327d3e1e57e2e7153b719ed985/src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntity_QuantumForceTransformer.java#L697) it looks like the "sum of halves" (`aChancePerOutput / 2 * (aOutputsAmount - 1)`) is directly assigned to the boosted output chance instead of adding to it. Same for the +1 tier case.

For the example above, the actual output chances are now:
33%/33%/33% -> **33%**/16%/16%
The sum of chances is less than 100%, and in this case it makes Neptunium focusing **strictly worse** than no focusing at all.

### Your Proposal

Fix the formulas for +0/+1 tiers by changing `=` to `+=`.

### Final Checklist

- [ ] I have asked other people and they confirm they also have this problem by follow the exact steps I described above, or this does not need reproducing, e.g. recipe loophole.

miozune:
Close https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/15082

### DIFF
--- a/src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntity_QuantumForceTransformer.java
+++ b/src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntity_QuantumForceTransformer.java
@@ -694,7 +694,7 @@ public class GregtechMetaTileEntity_QuantumForceTransformer
                 for (int i = 0; i < tChances.length; i++) {
                     if (doNeptunium) {
                         if (i == aChanceIncreased) {
-                            tChances[i] = aChancePerOutput / 2 * (aOutputsAmount - 1);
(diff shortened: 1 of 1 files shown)


Pelotrio (approved):
Interesting find and correct code change

PR #815, opened 2023-12-27
Thermal Boiler rework

### LINKED ISSUE #14540: GT++ Thermal Boiler ignores voiding output setting.

abdiel_kavash

### Your Pack Version

2.4.0

### Your Server

SP

### Java Version

Java 19

### Type of Server

Single Player

### Your Expectation

Build the Thermal Boiler multiblock, set it to not void fluids or items via GUI. Wait until its output bus/hatch fills up. (Or build the multi without one.)

I expect the boiler to stop operation when the output is full according to UI setting.

### The Reality

Thermal Boiler keeps running and voiding output items/fluids regardless of UI setting.

### Your Proposal

Either make the Thermal Boiler respect the voiding mode setting, or if this is intended disable the UI element (like for example turbines or LHE).

### Final Checklist

- [ ] I have asked other people and they confirm they also have this problem by follow the exact steps I described above, or this does not need reproducing, e.g. recipe loophole.

### LINKED ISSUE #14541: GT++ Thermal Boiler produces Tinker's Construct bronze.

abdiel_kavash

### Your Pack Version

2.4.0

### Your Server

SP

### Java Version

Java 19

### Type of Server

Single Player

### Your Expectation

GT++ Thermal Boiler, when processing Pahoehoe Lava, outputs Electrum, Tantalum, and Bronze nuggets (as well as obsidian and tungstate dust). My intention was to feed all the nuggets into an Alloy Smelter to turn them into ingots.

### The Reality

Electrum and Tantalum are produced as GT-compatible nuggets, however Bronze is output as Tinker's Construct nuggets. These can not be processed to ingots in any machine other than the smeltery.

### Your Proposal

Make the boiler output GT-compatible bronze nuggets. A hard IV-tier machine should not be making primitive tier of metals.

### Final Checklist

AbdielKavash:
The main contribution of this PR is to rework the Thermal Boiler to use `ProcessingLogic` and true RA recipes, instead of ad-hoc hardcoded processing. Along with this, there is a number of small changes and improvements:

* All recipes now return correct GT oredict items. (Fixes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/14541.)

* Void protection for both fluids and items now works correctly. (Fixes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/14540.)

* For lava and Pahoehoe lava, recipe time has been increased from 1 tick to 1 second to reduce recipe checks. All inputs and outputs have been increased proportionally. (*Note: If you compare NEI recipes, you might think that the output has been significantly nerfed. This is not true. In the current build, the NEI diagram *lies* and shows probabilities 10 times higher than the machine actually outputs. This has been fixed and the NEI diagram is now correct. The average amounts of outputs actually produced have not been changed.*)

* Previously, a lava filter would lose one point of durability every 30 seconds of machine operation. This was slightly buggy (one could pause processing for a tick at the right moment to skip this check), and is incompatible with the current processing logic. Now, the filter has a 1/30 chance of losing one point of durability with every operation. On average, this means that one filter will last the same amount of time.

* Updated the NEI diagram:

Now you can see ~how much~ how little steam this makes, without having to look it up on the wiki!

* The progress screen now shows all outputs, including cooled fluid and steam:

(*Note: There is currently a bug where the steam output rate does not update while the GUI is open. This is purely a display issue, the actual amount of steam produced correctly increases as the boiler heats up. I have not been able to track this down.*)

* Even though the NEI diagrams only show regular water as possible coolant, the boiler can still use Distilled Water as before. However this should probably be deprecated. Just use a reservoir hatch.

* When the boiler runs out of water, instead of exploding instantly, there is now a short grace period to refill it. This was inspired by the Large Heat Exchanger. (No steam is produced while no water is present.)

* As far as I know, no balancing factors have been changed (other than RNG variance due to longer recipe times), and all existing setups should continue working exactly as before without the need for any changes.


Steelux8 (changes requested):
I appreciate the changes to this multi, I set up automation for it and did notice the amount of outputs wasn't what I expected. However, since it's being changed, I'll request the removal of all nuggets and small dusts as outputs, to be replaced with ingots and normal dusts.

To do this, either multiply inputs and recipe time by 9 and keep chances for the nuggets to ingots, or adjust the chances while changing nuggets and dusts. It's fine if you approximate chances in this change, this multi is optional, not very popular and isn't going to affect balance with slight approximations.

Dream-Master:
I would not add ingots just dust .

Steelux8:
Does it make more sense for the metals to come out as dust or ingots? I'm not really sure myself, maybe dust is better.

AbdielKavash:
How about doing something different and making them come out as molten fluids? That would make automation a little bit more interesting (you need separate output hatches and solidifiers, or somehow ration the fluid in 144mB increments). For something that is legitimately this powerful, maybe adding an interesting automation challenge could be good.

Dream-Master:
molten is possible too. Not sure if it can set in % chance (fluids)

Dream-Master:
@AbdielKavash any news on this pr ?

AbdielKavash:
OK, so I did a bit more on this than I originally intended. I have reworked the existing recipes, slightly buffed power production, and added one new and one compatibility recipe. Details below:

### General

All recipes now run for 1 second. Lava and Pahoehoe Lava recipes now consume an even 1000 L per operation (per second). Output amounts have been adjusted to match (see details below). If the player is generating lava from an infinite source (Everburn Urn, Serenade of the Nether, ...) this shouldn't make a significant difference. But this is a small buff to players pumping lava from a finite source, as the average output *per bucket of lava* is slightly increased across the board.

### Power Generation

The Thermal Boiler (TB) here competes with the Large Heat Exchanger (LHE). Both can use hot fluids to generate steam. However the TB, despite being a tier and a half higher (casings sneakily require the Alloy Blast Smelter, which is mid-IV), is woefully inadequate in this contest. This is, of course, understandable, as the primary purpose of the TB is resource generation, not power. Regardless, I have buffed the steam output for both types of Lava.

The TB now produces 40,000 Steam per second, which translates to 1,000 EU/t at 100% turbine efficiency. For comparison, a LHE produces up to 4,000 EU/t with regular steam or 12,000 EU/t with superheated steam.

Existing recipe for Solar Salt has been adjusted, and a new recipe for IC2 Hot Coolant has been added. These recipes produce no items, and have been balanced for parity with the LHE. Both recipes produce superheated steam at half the rate of the LHE, while also consuming fluids at half the speed. Thus the EU per unit of fuel is the same, but the TB produces power slower. (12,000 EU/t compared to the LHE with 24,000.)

A new recipe for Blazing Pyrotheum has been added, which also produces power. See below for details.

### Item Generation

The main goal of the TB is to passively provide a slow trickle of basic and medium-tier materials. It should not be able to directly compete with more typical passive sources in the same tier (crops, bees, etc.) and neither with automated mining. Should more parity be desired, output can easily be buffed further. For balancing, I have referenced both the previous TB recipes, and centrifuge lava recipes. The yield per time of the new TB is comparable or slightly higher than the centrifuge, however the TB consumes lava at a *much* faster rate. (Centrifuge uses 25-50 lava/sec, TB uses 1,000 lava/sec.)

I have removed all the nuggets and small dusts from the old recipes, and replaced them with **impure piles of dust**. You wouldn't expect scraps of molten rock that get caught on a filter to have a quality ready for processing, right? Average output per material is slightly lower than before; however this is offset by the fact that centrifuging the impure piles can yield extra byproducts.

Materials output from both types of lava have been slightly shuffled around, now including a few types of igneous rock. For Pahoehoe Lava, this includes Sky Stone (AE2), which previously had no passive source other than (slow) TC Crucible alchemy. See table below for full details.

Obsidian output has been nerfed, especially for pahoehoe lava. It was produced at such a dumb rate that it was only trashing the output system.

A lava filter is still required (and slowly used up) to get any outputs other than obsidian, the mechanic or balance has not been changed from the initial PR.

A new recipe has been added, using Blazing Pyrotheum, which produces more exotic materials. See below for details.

### Pyrotheum Processing

I was thinking of what other hot fuels would make sense for the TB, and thought of Blazing Pyrotheum. So I added a recipe for it. Pyrotheum can't be simply pumped out of a hole in the ground somewhere, it has to be mixed from several basic materials. It can be obtained from bees, but the centrifuging recipe for the comb has a massive EU cost, and will likely not be fast enough to keep up with the TB without either overclocking or heavy parallelization.

As a reward for supplying 100 L of pyrotheum per second, the player is rewarded with 6,000 EU/t worth of superheated steam, and even more rare materials: this time, from the platinum group line. (Once again, in the form of impure dusts, and still requiring a lava filter.)

The EU/t output means that the whole line can be made power positive (depending on how exactly one sources the coal, blaze powder, etc.) It takes somewhere around 2,000 EU/t to make enough pyrotheum to keep up with the boiler, using the mixer recipe. On the other hand, with bees, this is practically impossible due to the comb centrifuging step, even with the GT++ centrifuge.

The platline outputs follow the same philosophy as everything else: they will be a steady passive trickle, not nearly enough to keep up with demand in IV+. I was in fact overly cautious when balancing this, I would be open to buffing these outputs.

Instead of obsidian, the pyrotheum recipe outputs blocks of coal. Taking into account the coal dust used to make the liquid, the loop is slightly coal-positive. This is intentional; coal here serves as an extra byproduct in a round-about way.

The last output is Fiery Steel (from TF). This is mostly a thematic "useless" byproduct, since it doesn't really make sense for pyrotheum to have stone in it; but it can be tossed into a Magic Energy Converter for some extra power (about 2,000 EU/t extra in total).

### Balance Details

See table for exact ratios of outputs. All outputs, unless otherwise specified, are in the form of Impure Pile of (X) Dust.

#### Lava Processing (input: 1000 L/sec)

| Output | Chance | Avg. output per hour | Notable byproducts |
| --- | --- | --- | --- |
| Copper | 3% | 108 | Cobalt (12/h) |
| Tin | 2% | 72 | Iron (8/h) |
| Silver | 1% | 36 | Lead (4/h) |
| Gold | 0.8% | 28.8 | Copper (3.2/h) |
| Redstone | 1.5% | 54 | |
| Black Granite | 4% | 144 | Fluorine (2,618 L/h) |
| Basalt | 4% | 144 | |
| Obsidian (Block) | 5% | 180 | |
| Pahoehoe Lava | | 1,000 L/sec | See table below |
| Steam | | 40,000 L/sec | 1,000 EU/t |

#### Pahoehoe Lava Processing (input: 1000 L/sec)

| Output | Chance | Avg. output per hour | Notable byproducts |
| --- | --- | --- | --- |
| Electrum | 0.5% | 18 | Gold (2/h) |
| Tantalite | 3% | 108 | Tantalum (24/h) |
| Bauxite | 1.5% | 54 | |
| Ilmenite | 2% | 72 | Titanium (14.4/h) |
| Tungstate | 1% | 36 | Tungsten (5.14/h) |
| Phosphate | 1% | 36 | Phosphorus (11.2/h) |
| Sky Stone (Block) | 4% | 144 | |
| Obsidian (Block) | 5% | 180 | |
| Steam | | 40,000 L/sec | 1,000 EU/t |

#### Blazing Pyrotheum Processing (input: 100 L/sec)

| Output | Chance | Avg. output per hour | Notable byproducts |
| --- | --- | --- | --- |
| Thorianite | 2% | 72 | Thorium (32/h) |
| Scheelite | 1.5% | 54 | Tungsten (9/h) |
| Platinum Metallic Powder | 1.5% | 54 | Platinum (38.5/h), other Pt group metals in trace amounts |
| Rarest Metal Residue Dust | 1% | 36 | Iridium (18/h), Osmium (1.8/h) |
| Iridium Metal Residue Dust | 0.5% | 18 | Iridium (18/h) |
| Fiery Steel Dust | 2% | 75 | |
| Coal (Block) | 5% | 180 | Coal Dust (1,620/h); note that making pyrotheum for the boiler consumes 1,440 coal dust/hour; for a net 180 coal dust/hour |
| Pahoehoe Lava | | 1000 L/sec | See table above |
| Superheated Steam | | 80,000 L/sec | 6,000 EU/t |

-----------

I am open to suggestions or feedback on any of the above. I had some more plans for the machine (for example, QFT-like focusing on specific inputs using a programmed circuit), but that can wait for later. This started as a polish/bug fix, and turned into a full rework already.

Steelux8:
I looked at the rework description in the last comment. First off, thank you very much for documenting the new values from your changes, and removing the nuggets as I requested. I would have approved this PR right away if the nuggets were changed to ingots/dusts without changing the final output rate, but this rework is something more than that.

I didn't want to rework this multi because, frankly, it's an old one whose purpose within GTNH quickly became to maintain the old lava centrifuge recipes, within this specific machine with the lava filter mechanic. As you noted, it was very weak for steam powergen and, honestly, I think the new values don't change that. I doubt anone would set these up for their main powergen, nor do I think the Thermal Boiler should be competing directly with the LHE for this purpose.

The two lavas are pretty similar to how they were before, although it is annoying to require oreproc to get the same outputs, although more realistic. The main gripe I have with this is the Blazing Pyrotheum processing addition, which I would prefer to not have because there are several problems with it:

- Where are these platline materials coming from, and the thorium/tungsten-related dusts? I don't remember this Pyrotheum having any of these things in it;
- 100L/s input rate is equal to 10 Volcanuses. Even if it's not a difficult fluid to make, that rate is still far too high per Thermal Boiler, unless the reward was fantastic, which is my next point;
- The average outputs of the important dusts are around 1/min, which is terrible compared to the usual methods of getting these dusts in IV. This won't break balance by being added, but I see little point in adding the Blazing Pyrotheum processing if it will be irrelevant like this.

AbdielKavash:
Fair enough. My idea for the pyrotheum recipe was for it to be something small and optional, that one could set up for fun on the side. In fact, I mathed it out more closely, and you can set up a fully self-contained system with an EEC sourcing the materials for the pyrotheum, and run the whole thing on the steam the TB makes. The entire system will simply produce the outputs indefinitely without the need for any additional inputs. I think that's something that would be fun (at least, that is the kind of thing that *I* personally find fun), even though not particularly "good".

As I wrote above, I set the output rates intentionally low. Because this closed loop system is possible, if the output was meaningful, people could just spam this setup a bunch of times, and completely avoid having to mine anything that the machine produces. In fact, speaking from experience, this is entirely possible with the existing lava recipes already; which matters particularly for tantalum and tungsten.

I do get the realism point; if you think of pyrotheum purely as a mixture of molten blaze/redstone/sulfur/coal, and not as a "magical hot liquid", then it does not make sense for it to start producing extra outputs. But then you can also ask where do all the impurities come from in water that was synthesised from hydrogen and oxygen, or magnesium in obsidian generated from redstone, and so on. Nevertheless, I accept this as a potential problem.

I thought the change to impure dusts would make the processing a little more interesting, though I don't have a strong opinion on it either way. I can revert it to just normal dusts if you think that'd be better.

src/main/java/gtPlusPlus/core/recipe/RECIPES_GREGTECH.java
```diff
@@ -1101,39 +1102,85 @@ private static void distilleryRecipes() {
...
+                        GT_OreDictUnificator.get(OrePrefixes.dustImpure, Materials.Bauxite, 1),
+                        GT_OreDictUnificator.get(OrePrefixes.dustImpure, Materials.Ilmenite, 1),
+                        GT_OreDictUnificator.get(OrePrefixes.dustImpure, Materials.Tungstate, 1),
+                        GT_OreDictUnificator.get(OrePrefixes.dustImpure, Materials.Phosphate, 1),
+                        ItemUtils.getItemStackFromFQRN("appliedenergistics2:tile.BlockSkyStone", 1),
+                        new ItemStack(Blocks.obsidian, 1, 0))
+                .outputChances(50, 300, 150, 200, 100, 100, 400, 500).specialValue(1).duration(1 * SECONDS).eut(0)
+                .addTo(thermalBoilerRecipes);
+
+        /*
```
chochem: why is this commented out? if you dont want it, remove it please.

Dream-Master: maybe we run it on zeta first and see if it is balance and remove this after a test ?

Pxx500:
Overall the PR looks very good but I'd consider change a few things:

1) Low tier outputs like iron, copper etc. should be increased so they are at least more common then the rare ones so it doesn't seem weird that usually rare resources are magically more common then common ones.
2) Alumina could be added to the lavas since aluminium is one of the main lava components.
3) Blazing pyrotheum has a determined formula, and the recipe proposed is essentially magically turning one resource into another. I propose to add a "dummy" cobblestone input so that it looks as if the pyrotheum is melting cobble and extracting resources from it (yes the stone can be centrifuged for other resources but it's at least a little bit more plausible)
4) Lower/Increase the amount of tantalite (I'm not really sure about this one, but my logic is that tantalum is an uncommon resource that significantly boosts your circuit component production so it should stay rare and 24/hr is a lot, but considering that this multi is mid-IV tier and the resource costs ramp up significantly it might be a good multi to give the player a stable tantalum production methods thus requiring this number to be even higher though 24/hr is not small so I suppose this is tier dependent)
5) The provided tables are not sorted by rarity so I hope in game the outputs will be

AbdielKavash:
> why is this commented out? if you dont want it, remove it please.

I am sorry, honestly I don't know what I was thinking. It makes no sense to leave it in like that. I will remove completely until there is more discussion about the balance.

> Low tier outputs like iron, copper etc. should be increased so they are at least more common then the rare ones so it doesn't seem weird that usually rare resources are magically more common then common ones.

That is already the case though? Unless you mean the higher tier resources like Ilmenite, Tungstate, etc, which are balanced around the rate of the final product (Titanium, Tungsten).

> Alumina could be added to the lavas since aluminium is one of the main lava components.

Good idea, it should probably replace Bauxite (or Redstone from normal lava).

> Blazing pyrotheum has a determined formula, and the recipe proposed is essentially magically turning one resource into another. I propose to add a "dummy" cobblestone input so that it looks as if the pyrotheum is melting cobble and extracting resources from it (yes the stone can be centrifuged for other resources but it's at least a little bit more plausible)

Pyrotheum also contains blaze dust, which can't be synthesised out of anything other than mob drops or crops. Who knows what impurities are in that? I find that about as "plausible" as getting rare materials out of cobblestone that magically appears out of nothing.

> Lower/Increase the amount of tantalite

The amount of tantalite has actually been *nerfed* from the previous version, as it used to be produced from both lava and pahoehoe lava. The 2.5.1 setup produces 40 tantalum/h total; the proposed one only 24/h after processing the tantalite.

> The provided tables are not sorted by rarity so I hope in game the outputs will be

They are sorted by the logical order of tier progression; with the "filler" stone last. Since the more complex resources are balanced around the final output rate, not the raw material, the percentages seem off. But this can be easily changed.

---

I will, for now, revert exactly the rates and outputs (including steam) from current stable version; only replacing nuggets with ingots. If some consensus is reached on the new outputs/new lines, it is very easy to simply change the recipes again.

AbdielKavash:
New recipes:




Comparison of outputs after some non-specific time of running both simultaneously:

2.5.1:

New:


Keep in mind that the new version has a much greater variance, because of both the replacement of nuggets with ingots, and the increase of processing time to 20 ticks.

Dream-Master:
@Steelux8

Steelux8 (approved):

PR #818, opened 2023-12-30
GT++ Fish Catcher Water Block Check

bitonality:
The existing code for adjacent water blocks/fish catchers is logically broken and non-performant. 

### Proposal

- Remove the string contains from the adjacent block check. It hurts performance and allows for any block containing "Water" to count as legal blocks (e.g. Chisel "Waterstone").
- Fix the internal logic such that fish catchers logically require two+ water blocks touching faces. This brings the internal logic up to parity with the tooltip's description of the intended behavior.
- Fix the scaling of loot acquisition such that more adjacent water+fish catchers leads to faster loot acquisition. I implemented what I believe is the intended behavior based on the existing code.

> 0 Adjacent Blocks | 0 ticks (No-op)
> 1 Adjacent Blocks | 0 ticks (No-op)
> 2 Adjacent Blocks | 6800 ticks
> 3 Adjacent Blocks | 5600 ticks
> 4 Adjacent Blocks | 4400 ticks
> 5 Adjacent Blocks | 3200 ticks
> 6 Adjacent Blocks | 1750 ticks

### Performance Notes

I wouldn't read too much into the following profiling, but it at least seems to confirm that removing the string contains improves performance (as expected).

## Release Implementation - 50 Fish Catchers
GT++ Fish Catcher Original

## Proposed Implementation - 120 Fish Catchers
GT++ Fish Catcher Update

### DIFF
--- a/src/main/java/gtPlusPlus/core/tileentities/general/TileEntityFishTrap.java
+++ b/src/main/java/gtPlusPlus/core/tileentities/general/TileEntityFishTrap.java
@@ -29,7 +29,7 @@ public class TileEntityFishTrap extends TileEntity implements ISidedInventory {
     private final InventoryFishTrap inventoryContents;
     private String customName;
     // The number of water blocks is used as an index to get the tick rate.
-    private final static short[] waterBlocksToTickRate = new short[] { 0, 0, 6800, 5600, 4400, 3200, 1750 };
+    private final static short[] waterBlocksToTickRate = new short[] { 0, 0, 5600, 4400, 3200, 2000, 1750 };
     private int surroundingWaterBlocks = 0;
 
     public TileEntityFishTrap() {
--- a/src/main/java/gtPlusPlus/core/tileentities/general/TileEntityFishTrap.java
+++ b/src/main/java/gtPlusPlus/core/tileentities/general/TileEntityFishTrap.java
@@ -29,8 +29,7 @@ public class TileEntityFishTrap extends TileEntity implements ISidedInventory {
     private final InventoryFishTrap inventoryContents;
     private String customName;
     // The number of water blocks is used as an index to get the tick rate.
-    // private final static short[] waterBlocksToTickRate = new short[] {0, 0, 6800, 5600, 4400, 3200, 1750};
-    private final static short[] waterBlocksToTickRate = new short[] { 0, 0, 100, 80, 60, 40, 20 };
+    private final static short[] waterBlocksToTickRate = new short[] {0, 0, 6800, 5600, 4400, 3200, 1750};
     private int surroundingWaterBlocks = 0;
 
     public TileEntityFishTrap() {
--- a/src/main/java/gtPlusPlus/core/tileentities/general/TileEntityFishTrap.java
+++ b/src/main/java/gtPlusPlus/core/tileentities/general/TileEntityFishTrap.java
@@ -26,19 +26,20 @@ import gtPlusPlus.core.util.minecraft.ItemUtils;
 public class TileEntityFishTrap extends TileEntity implements ISidedInventory {
 
     private int tickCount = 0;
-    private boolean isInWater = false;
     private final InventoryFishTrap inventoryContents;
     private String customName;
-    private int waterSides = 0;
-    private int baseTickRate = 600 * 5;
+    // The number of water blocks is used as an index to get the tick rate.
+    // private final static short[] waterBlocksToTickRate = new short[] {0, 0, 6800, 5600, 4400, 3200, 1750};
+    private final static short[] waterBlocksToTickRate = new short[] { 0, 0, 100, 80, 60, 40, 20 };
+    private int surroundingWaterBlocks = 0;
 
     public TileEntityFishTrap() {
         this.inventoryContents = new InventoryFishTrap();
     }
 
-    private boolean isSurroundedByWater() {
+    private int getNumberOfSurroundingWater() {
         if (!this.hasWorldObj() || this.getWorldObj().isRemote) {
-            return false;
+            return 0;
         }
         final Block[] surroundingBlocks = new Block[6];
         surroundingBlocks[0] = this.worldObj.getBlock(this.xCoord, this.yCoord + 1, this.zCoord); // Above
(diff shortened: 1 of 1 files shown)


src/main/java/gtPlusPlus/core/tileentities/general/TileEntityFishTrap.java
```diff
@@ -50,23 +50,18 @@ private boolean isSurroundedByWater() {
...
-        if ((waterCount >= 2) && (trapCount <= 4)) {
-            int aCheck = trapCount + waterCount;
-            this.waterSides = MathUtils.balance(aCheck, 0, 6);
-            return true;
+        // Explicitly check for at least 2 water blocks.
+        if (waterCount < 2) {
+            return 0;
         }
-        return false;
+        // Only allow the first four traps to could towards valid neighbor blocks.
```
Glease: typo
```suggestion
        // Only allow the first four traps to count towards valid neighbor blocks.
```

src/main/java/gtPlusPlus/core/tileentities/general/TileEntityFishTrap.java
```diff
@@ -26,19 +26,19 @@
 public class TileEntityFishTrap extends TileEntity implements ISidedInventory {
 
     private int tickCount = 0;
-    private boolean isInWater = false;
     private final InventoryFishTrap inventoryContents;
     private String customName;
-    private int waterSides = 0;
-    private int baseTickRate = 600 * 5;
+    // The number of water blocks is used as an index to get the tick rate.
+    private final static short[] waterBlocksToTickRate = new short[] { 0, 0, 5600, 4400, 3200, 2000, 1750 };
```
Glease, line 32: 6800 is gone and you inserted a completely new value 2000. Why?

Glease (dismissed):
besides the other two comment, maybe i'm being dumb here, but I don't see how original logic is broken? mind to explain a bit?

src/main/java/gtPlusPlus/core/tileentities/general/TileEntityFishTrap.java
```diff
@@ -50,23 +50,18 @@ private boolean isSurroundedByWater() {
         int waterCount = 0;
         int trapCount = 0;
         for (final Block checkBlock : surroundingBlocks) {
-            if ((checkBlock == Blocks.water) || (checkBlock == Blocks.flowing_water)
-                    || checkBlock.getUnlocalizedName().toLowerCase().contains("water")
-                    || (checkBlock == ModBlocks.blockFishTrap)) {
-                if (checkBlock != ModBlocks.blockFishTrap) {
```
bitonality, line 56: @Glease 
The broken logic stems from the fact that waterCount and trapCount aren't discretely compared and are pre-aggregated by incrementing water AND trap count when a trap block is present. This leads to the case where:
- 0 water blocks adjacent
- 2 trap blocks adjacent
would be internally tracked as:
- 2 water blocks
- 2 trap blocks

This is in addition to the string.contains("water") check which allows unintended blocks to count as water blocks and hurts performance by having to do a string.contains() six times per trap per 20 ticks.

src/main/java/gtPlusPlus/core/tileentities/general/TileEntityFishTrap.java
bitonality: I got some feedback in the mod channel on the discord to adjust these values. 
Upon reflecting, I will revert these to the intended values and any future re-balancing can be addressed in a separate change with the appropriate people involved in the balance discussion. Apologies for jumping the gun and going out of scope with this specific change.

src/main/java/gtPlusPlus/core/tileentities/general/TileEntityFishTrap.java
Glease: I see what you mean. Sorry for overlooking this chunk of the diff. However, I'm afriad this is intentional. I have sent a DM to alk (original author) and she confirms my theory. Fish catcher are also filled with water, so it counts as water.

src/main/java/gtPlusPlus/core/tileentities/general/TileEntityFishTrap.java
Glease: Well I have completely no objection over this trivial balance changes, and it's probably not going to be objected by other people either. However you should clearly state in PR description that you did so, instead of just hiding this change among other things

Glease (commented):
We can probably pick up the no string contains change as an optimization. Spring water and alike are too rare and that's not worth the cpu cycles. However, the others are pure balance changes and are therefore up to discussion. I'm only doing the **code** review, whether the content is needed will have to be handled separately.

src/main/java/gtPlusPlus/core/tileentities/general/TileEntityFishTrap.java
bitonality: Thank you for clarifying the original intent! I will change the scope of this PR to only remove the string.contains and preserve the intended logic. Do you think it's worth updating the tool-tip to more accurately communicate the intended behavior? Proposing something like the following:

Current:
  list.add("This trap catches fish faster if surrounded by more water blocks");
  list.add("Can also be placed beside upto 4 other fish traps");
  list.add("Requires at least two faces touching water");
  list.add("1/1000 chance to produce triple loot.");
 
Proposed
  list.add("Requires at least two faces touching water or other fish catchers");
  list.add("Rates improve when surrounded by water");
  list.add("Rates also improve when surrounded by other fish catchers (max of four)");
  list.add("1/1000 chance per catch to produce triple loot.");

Steelux8 (approved):
I'll approve these changes, based on the fact that the baseTickRates were changed as follows:

6800, 5600, 4400, 3200, 1750
to
5600, 4400, 3200, 2000, 1750

It's a trivial balance change that probably won't matter for anything, but I'll acknowledge it here so that it can be reverted if there are any problems.

Steelux8:
Closes [#15249](https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/15249).

PR #820, opened 2024-01-05
Add a tooltip for QFT Fluid mode

BlueWeabo:
Adds a tooltip to the qft to say what Fluid Mode is and how to activate it


AbdielKavash (commented):
Just a tiny bit of pedantry, usually tool tooltips are written as "Use (tool) to (do thing)". So here it would be "Use screwdriver to enable Fluid mode".

Normally I wouldn't comment on such small wording things, but since the entire PR is about changing the tooltip I couldn't resist :-)

BlueWeabo:
No problem with changing it

bombcar (approved):

PR #821, opened 2024-01-05
Fix chemical plant

HoleFish:
- Supports crib
- Correctly gets catalyst from catalyst housing


chochem:
what was the problem with catalysts?

HoleFish:
you can install a catalyst housing and put the catalyst into input bus to run chemical plant. It should not.

BlueWeabo (approved):

PR #825, opened 2024-01-12
Advanced Muffler Hatch improvements

AbdielKavash:
This PR contains several improvements to Advanced Muffler Hatches:

* As of 2.5.x, if the advanced muffler has no filter or the filter is depleted, the multiblock will immediately halt operation and void. Now, an advanced muffler can continue operating even without an air filter, but it will only reduce pollution by as much as a basic muffler of the same tier would.

* When the air filter is depleted, the advanced muffler will now try to pull a new filter from an input bus of the multiblock. This simplifies automation, especially for multiblocks which force the muffler to be in a position where none of its sides are accessible (for example, EBF, Pyrolyze Oven, Large Combustion Generator, ...)

* Pollution output is now fixed at 95% for LV, and 0% for UHV advanced mufflers. (Previously, 100% and 2.81% respectively.) This means that the advanced muffler is now always better (or at least not worse) than a basic muffler. The formula for other tiers is not changed.

* The tooltip no longer mentions an exhaust gas recovery amount that is different from the pollution filtering amount. This never worked in the first place; and is in fact fundamentally impossible to do with the way the EBF calculates exhaust amounts. (I couldn't find a github issue for this, but I remember somebody mentioning it in discord.)

* Cleaned up some obsolete/unreachable code.

There is currently one small inconvenience, that is if you build a multiblock with an advanced muffler, and place the air filter in an input bus, the filter will only be retrieved the first time the multiblock vents pollution. This means that specifically for the EBF, the amount of exhaust gas will be calculated as if using a basic muffler before the first pollution cycle. I don't believe this is possible to fix without changing base multiblock code. If you really care about a few % worth of exhaust gas, place a filter in the muffler hatch manually; it will be replaced correctly from that point on.

The removed code in `LargeRocketEngine.java` performed the same function that the mufflers now do by themselves (replacing a depleted filter), and thus is now redundant.

### DIFF
--- a/src/main/java/gtPlusPlus/core/item/general/ItemAirFilter.java
+++ b/src/main/java/gtPlusPlus/core/item/general/ItemAirFilter.java
@@ -88,6 +88,10 @@ public class ItemAirFilter extends Item {
         return true;
     }
 
+    public static final long getFilterMaxDamage(final ItemStack aStack) {
+        return aStack.getItemDamage() == 0 ? 50 : 2500;
+    }
+
     public static final long getFilterDamage(final ItemStack aStack) {
         NBTTagCompound aNBT = aStack.getTagCompound();
         if (aNBT != null) {
@@ -119,7 +123,7 @@ public class ItemAirFilter extends Item {
             createNBT(stack);
         }
         double currentDamage = getFilterDamage(stack);
-        double meta = stack.getItemDamage() == 0 ? 50 : 2500;
+        double meta = getFilterMaxDamage(stack);
         double durabilitypercent = currentDamage / meta;
         return durabilitypercent;
     }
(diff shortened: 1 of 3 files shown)


Glease:
this seems like a much more sensible approach towards fixing adv muffler.

chochem:
I mean this certainly works.

But any particular reason why the adv. muffler has a exponential formula (x0.64 each tier), while the normal one has a linear one (-12.5% per tier)? Another option could just be to take that linear formula and shift it by -lets say- one tier to make the advanced better.

Glease:
Because it always have been exponential until bart changed normal ones to be linear. obviously bart isn't going to care about gt++ and even if he does, gt++ wasn't LGPL back then

chochem:
I would call that  _no particular reason_ - just some history XD

AbdielKavash:
I'm not particularly attached to any formula, I simply left the one that was already there unchanged. If anything, they could be buffed more in the first 1-2 tiers where the investment needed to get them going is the greatest. (No AE to autocraft/autostock filters.)

Although as long as the Air Filter multiblock remains in its current state, it will always be an objectively better way of dealing with pollution; and Advanced Mufflers are going to have only extremely niche uses at all tiers, no matter how effective they are. But that's for another discussion.

Steelux8 (approved):
Looks good, thank you for the cleanup, fixes and the advanced muffler improvement. I've never used one of these, but they should be better than a regular muffler in all situations IMO.

PR #834, opened 2024-02-13
Enable perfect OC on the matter fabrication CPU recycler mode

GDCloudstrike:
Enables perfect OC for the recycler mode of the matter fabrication CPU.
Currently this mode is completely unusable on large scale as the recyclers are 1. way too slow and 2. far too power hungry to be worth it.
Enabling perfect OC for this mode fixes both of those issues and makes using UUA viable (granted one can supply the necessary items for scrap production)
image
https://docs.google.com/spreadsheets/d/1YxpXsn6DdMKQ2Kg9punc7zUSOACXb-F53B6kf6XLwTk/edit?usp=sharing

### DIFF
--- a/src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntity_MassFabricator.java
+++ b/src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntity_MassFabricator.java
@@ -284,7 +284,7 @@ public class GregtechMetaTileEntity_MassFabricator
     @Override
     protected void setupProcessingLogic(ProcessingLogic logic) {
         super.setupProcessingLogic(logic);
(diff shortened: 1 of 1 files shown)


Steelux8:
Is it necessary to buff replication? Last I heard, there was an agreement on replication being removed in favor of producing materials in their original, intended ways.

GDCloudstrike:
> Is it necessary to buff replication? Last I heard, there was an agreement on replication being removed in favor of producing materials in their original, intended ways.

This change does not have replication as its main target but rather the general use of UUM in crafting recipes. Replication does not use nearly as much UUM as crafting does, so increasing production wont directly buff replication as the multi already makes enough UUM for that (without this change). But it does cut down on the amount of matter fabrication CPUs needed in late/endgame while reviving a multiblock mode that was basically useless previously.

BlueWeabo (approved):
Go for it

PR #837, opened 2024-02-22
Add alternate recipes for neptunium and fermium plasma

GDCloudstrike:
Adds alternate higher tier recipes for neptunium and fermium plasma which are faster than their mk4 counterparts but also require part of the fusion chain, adding complexity in return.
These two plasmas are notoriously slow currently and the proposed alternate recipes aim to make their production less of a slog without making them free.

Neptunium:
image
Fermium:
image

### DIFF
--- a/src/main/java/gtPlusPlus/xmod/gregtech/loaders/recipe/RecipeLoader_ChemicalSkips.java
+++ b/src/main/java/gtPlusPlus/xmod/gregtech/loaders/recipe/RecipeLoader_ChemicalSkips.java
@@ -342,18 +342,18 @@ public class RecipeLoader_ChemicalSkips {
 
         // MK5 versions
         GT_Values.RA.addFusionReactorRecipe(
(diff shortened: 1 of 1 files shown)


HoleFish:
They are still MK4 recipes in compact fusions actually. "higher tier" is meaningless, just more complex :trollface:

GDCloudstrike:
> They are still MK4 recipes in compact fusions actually. "higher tier" is meaningless, just more complex :trollface:

That is a bug then, compacts should not skip fusion tiers

HoleFish:
A true MK5 recipe in compact fusions needs over 2.56G EU for startup. Currently it's not possible since special value is int

GDCloudstrike:
> A true MK5 recipe in compact fusions needs over 2.56G EU for startup. Currently it's not possible since special value is int

should be good now. Startup power is not the only limiting factor when it comes to fusion tiers <https://github.com/GTNewHorizons/GoodGenerator/pull/238>

BlueWeabo (approved):
I approve of this.

PR #839, opened 2024-02-23
Tree Growth Simulator can now harvest leaves and fruits, using appropriate tools.

AbdielKavash:
Inspired by https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/15513 and my own desire to have a method to farm fruit trees that is more scalable and less laggy then either Forestry multifarms or TC golems.

The Tree Growth Simulator multiblock can now farm all parts of a tree: logs, saplings, leaves, and fruits. To do this, it must be equipped with a proper tool: saw for logs, branch cutter for saplings, shears for leaves, and knife for fruit. Like before, more "advanced" versions of the tools (chainsaw, grafter, automatic snips...) increase output.

In order to be able to use different tools, the tools are now taken from an input bus, and the sapling needs to be placed in the controller slot. This way, multiple different tools can be used together to get all the different outputs at the same time. Previously, the item placement logic was reversed: the saw used to go in the controller slot, and saplings in an input bus. The new logic can recognize this setup and correctly swap the positions of items, without interrupting operation of existing setups.

Processing speed, wood output amount, energy use and scaling, and damage done to tools has not been changed from previous versions. Existing setups should continue running uninterrupted with the same specifications as before. Output amounts of other products scale with energy tier using the same formula, but are proportionally lower than wood output. (Nobody needs five thousand pineapples per second.)

Farming saplings no longer consumes fertilizer, instead, damage to the branch cutter tool is the resource cost. If a player used the TGS to farm saplings previously, they will need to insert a branch cutter to continue. I have briefly considered adding a slow rate of fertilizer consumption to all working modes; but this would break far more existing setups for no real reason.

Miscellaneous notes:
* Only the outputs that have a valid tool provided are generated. For example, if you only use a knife, only fruit will be produced. This way you don't need a separate setup to void logs/leaves if you don't want them.
* Very large trees (think redwoods, sacred oaks, and the like) now output more logs than before.
* Any fruit-bearing trees (vanilla oaks, Pam's, Forestry, etc.) produce their fruit when a knife is used.
* Added a few new trees that could not be farmed previously.
* Added farming of mushrooms. This is mostly so that there is an automated source of the cosmetic mushroom blocks; the EIG farms mushrooms much more efficiently and without needing a tool to do it.
* Forestry and Extra Trees saplings benefit from their genetic stats when farmed. (Part of this was already used before, I have expanded the functionality to stats affecting the drops of saplings and fruits.)
* Now uses `ProcessingLogic`.
* Redone the NEI recipe window to include all possible outputs. (See below.)
* Yes, the pocket multi-tool can be used in place of all the processing tools, this is intentional as a bit of an easter egg/discoverable feature. It has a production multiplier of 1 (lowest possible) and of course takes four times the durability damage when used this way.

Balance effects (all likely very minor):
* Large trees (redwoods and such) output more wood than before, up to 4x more for the largest ones. This can result in a truly stupid amount of wood when fed late-game power. However setups that rely on such amounts of wood (benzene of course) are not bottlenecked by the TGS anyway.
* Farming saplings now uses tool durability instead of liquid fertilizer. Both are about equally difficult to source, they just need a different setup. Does anyone need thousands of saplings anyway?
* Coconuts and some Forestry trees farmed in the TGS could be a very good new source of seed oil. Honestly I haven't run the numbers yet on how it compares to peanuts or pumpkins in the EIG. Maybe this could be a good biodiesel source? How relevant is that in IV+?

Potential future work:
* [x] Add/update questbook description (I will do this after this PR is accepted.)
* Localize new strings. (Localization keys are in, just need somebody to translate them.)
* Add more granularity to the output amount calculation so that different trees can be balanced more accurately.
* Add support for multiple different outputs in one mode (e.g., multiple different logs). This is a very low priority, there are only about 2-3 trees where this would be relevant.
* Ask somebody who actually knows Forestry tree breeding in depth to verify the genetics scaling values.
* If somebody can make a nicer progress bar for NEI that looks like a tree growing, I would love to add it.
* I would like at least one more efficient tool for harvesting fruit; currently the knife is the only option. But I can't think of anything else that would be appropriate. Maybe a Sense? That seems to be the closest thing, even though it doesn't really make sense.
* Maybe make the stats of the tool affect the speed/output somehow? Currently the only stat that matters is durability.

&nbsp;

<details>
<summary>Image gallery (click to expand)</summary>

:mangoman:

</details>


Csstform:
for fruit harvesting you could do something like this: https://www.starkbros.com/products/tools-and-supplies/tools-and-equipment/seymour-fruit-picker

AbdielKavash:
Right, I didn't want to add a whole new tool just for this. But if someone does the art for me, I might consider it!

It should also be easy to make it increase apple drops from leaves and such, when used in-world, to give it a purpose beyond just this.

Csstform:
picker
Nobody ever accused me of being a good pixel artist but this mostly looks like a picker if you want it.

src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntityTreeFarm.java
```diff
@@ -149,101 +131,53 @@ protected int getCasingTextureId() {
...
+    public int getMaxParallelRecipes() {
+        return 1;
+    }
 
-        if (this.lEUt > 0) {
-            this.lEUt = (-this.lEUt);
-        }
+    @Override
+    public boolean supportsBatchMode() {
+        // Batch mode would not do anything, processing time is fixed at 100 ticks.
```
BlueWeabo, line 165: You will want to also override isBatchModeEnabled and set it to false as people could've previously turned it on and they wouldn't be able to disable it.

src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntityTreeFarm.java
```diff
@@ -272,326 +206,574 @@ public IStructureDefinition<GregtechMetaTileEntityTreeFarm> getStructureDefiniti
...
+                }
+
+                ItemStack sapling = findSapling();
+                if (sapling == null) return CheckRecipeResultRegistry.NO_RECIPE;
+
+                EnumMap<Mode, ItemStack> outputPerMode = getOutputsForSapling(sapling);
+                if (outputPerMode == null) {
+                    // This should usually not be possible, outputs for all valid saplings should be defined.
+                    Logger.INFO("No output found for sapling: " + sapling.getDisplayName());
+                    return CheckRecipeResultRegistry.NO_RECIPE;
```
BlueWeabo: What you make a new RecipeResult so you could display that the sapling has no outputs defined instead of it saying no recipe?

src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntityTreeFarm.java
```diff
@@ -272,326 +206,574 @@ public IStructureDefinition<GregtechMetaTileEntityTreeFarm> getStructureDefiniti
...
+
+                    // Increase output by the relevant multipliers.
+                    ItemStack out = output.copy();
+                    out.stackSize *= tierMultiplier * modeMultiplier.get(mode) * toolMultiplier;
+                    outputs.add(out);
+                }
+
+                if (outputs.isEmpty()) {
+                    // No outputs can be produced using the tools we have available.
+                    return CheckRecipeResultRegistry.NO_RECIPE;
```
BlueWeabo: New CheckRecipeResult for no tools available for sappling

src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntityTreeFarm.java
```diff
@@ -272,326 +206,574 @@ public IStructureDefinition<GregtechMetaTileEntityTreeFarm> getStructureDefiniti
...
+    /**
+     * Attempts to find a tool appropriate for the given mode, and damage/discharge it by one use.
+     * 
+     * @param mode The mode to use. This specifies which tools are valid.
+     * @return Production multiplier based on the tool used, or -1 if no appropriate tool was found.
+     */
+    private int useToolForMode(Mode mode) {
+        for (ItemStack stack : getStoredInputs()) {
+            int toolMultiplier = getToolMultiplier(stack, mode);
+            if (toolMultiplier > 0) {
```
BlueWeabo: nit. can you reverse some of the ifs so that it goes with early exit instead? so this one would be `if (toolMultiplier < 0) continue;`

src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntityTreeFarm.java
```diff
@@ -272,326 +206,574 @@ public IStructureDefinition<GregtechMetaTileEntityTreeFarm> getStructureDefiniti
...
+    /**
+     * Get a list of possible outputs for a sapling, for each mode. This is either recovered from
+     * {@link #treeProductsMap}, or generated from stats of Forestry saplings.
+     *
+     * @param sapling A sapling to farm.
+     * @return A map of outputs for each mode. Outputs for some modes might be null.
+     */
+    private static EnumMap<Mode, ItemStack> getOutputsForSapling(ItemStack sapling) {
+        String registryName = Item.itemRegistry.getNameForObject(sapling.getItem());
+        if ("Forestry:sapling".equals(registryName)) {
```
BlueWeabo, line 550: nit. switch the places of what calls the equals. (I  know it works, but looks weird having that

src/main/java/gtPlusPlus/xmod/gregtech/loaders/recipe/RecipeLoader_TreeFarm.java
BlueWeabo, line 1: Any reason you have both `@Optional.Method` and the if check if the mod is loaded? Isn't only one enough?

src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntityTreeFarm.java
AbdielKavash: Done.

src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntityTreeFarm.java
AbdielKavash: Done.

src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntityTreeFarm.java
AbdielKavash: Good idea, done.

src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntityTreeFarm.java
AbdielKavash: Done.

src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntityTreeFarm.java
AbdielKavash: Just a habit. `"stringConstant".equals(stringVariable)` will not fail if `stringVariable` is null. Though here `registryName` should probably never be null anyway.

src/main/java/gtPlusPlus/xmod/gregtech/loaders/recipe/RecipeLoader_TreeFarm.java
AbdielKavash: No particular reason. Cleaned up the annotations.

BlueWeabo (approved):
Code wise is fine now. Someone with balance thoughts should chime in to make sure everything is okay.

src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/GregtechMetaTileEntityTreeFarm.java
BlueWeabo: I see. yeah that is understandable. can be left as is

PR #840, opened 2024-02-26
Free casings Matter Fabrication CPU

dahgul333:
now you need 36 casing (36<-40)
we need it for normal fit 8 Energy Hatches with Maintenance Hatch
 befor it you could only put 6 and without the Maintenance Hatch
6 was not enough somtimes so for prewenting random turn -offing utilized Machine Controller Cover , and with it and without Maintenance Hatch after some "Problems"   it went into an on/off cycle.
1)before :
image
2) after :
image


Dream-Master:
@dahgul333 overclock is fine but a 2X OC is op and not planned. Or what this change will doing ?

dahgul333:
> overclock is fine but a 2X OC is op and not planned. Or what this change will doing ?

8 energy hatches give a same speed as 6 here but with 8 and slot for Maintenance it will work more stable and adequate

@Dream-Master

koiNoCirculation:
@dahgul333 You dont have to do this, with UV Energy hatch you just need 5 energy hatches :);
63600b9bf29b330b852b567e820af9b5

dahgul333:
@koiNoCirculation
with 5 Umv it not work ...

Dream-Master (approved):
lets test this and if get overpowered we can revert it.

PR #841, opened 2024-02-27
Let remainingFlow become 125% of realOptFlow, rather than raw optimal…

koiNoCirculation:
… flow.

This can fix the inconsistance between optimal flow in tooltip and that in turbine.

When I try to use duranium huge turbine on my XL SC Turbine, I found that the actual output and steam consumption is far less than the calculation.
93449a920eca9b8cf737b1ae74245a13
XL SC turbine outputs 6768000EU/t on server, but according to calculation, output is parallelism * optimal flow * efficiency%  * fuel value of SC steam. The optimal flow in XL SC turbine with duranium huge turbine inserted and fast mode enabled is 48 * single optimal flow = 4915200mb/t. The turbine efficiency is 235%, and fuel value of SC steam is 100EU/mb. So final output should be 4915200 * 2.35 * 100 = 115507200 EU/t.  Obviously it doesn't work as the description on turbine tooltip.

Through reading the code I found that 6768000 = 6400 * 1.25 * 2.35 * 100 * 48 * 0.75 (effiency when flow over optimal).
While 6400 is the original optimal flow of the turbine, and has different multipliers when used in different type of machine.
I found that the  variable realOptFlow uses that multiplier but remainingFlow doesn't, so the optimal flow is limited to original flow, rather than actual flow. Through this change I let the steam/HP/SC turbine behave as tooltip on turbine describes.

Output of SC turbine with duranium huge turbine after fix:
de5c7bcfb085963fdbf7739c8ed4bc16

### DIFF
--- a/src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/turbines/GT_MTE_LargeTurbine_SCSteam.java
+++ b/src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/turbines/GT_MTE_LargeTurbine_SCSteam.java
@@ -61,7 +61,9 @@ public class GT_MTE_LargeTurbine_SCSteam extends GregtechMetaTileEntity_LargerTu
         int flow = 0;
         // Variable required outside of loop for
         // multi-hatch scenarios.
-        this.realOptFlow = (double) aOptFlow * (double) flowMultipliers[0];
+        this.realOptFlow = aOptFlow;
+        // this.realOptFlow = (double) aOptFlow * (double) flowMultipliers[0];
+        // Will there be an multiplier for SC?
         int remainingFlow = MathUtils.safeInt((long) (realOptFlow * 1.25f)); // Allowed to use up to
         // 125% of optimal flow.
 
--- a/src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/turbines/GT_MTE_LargeTurbine_SCSteam.java
+++ b/src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/turbines/GT_MTE_LargeTurbine_SCSteam.java
@@ -59,8 +59,9 @@ public class GT_MTE_LargeTurbine_SCSteam extends GregtechMetaTileEntity_LargerTu
         int tEU = 0;
         int totalFlow = 0; // Byproducts are based on actual flow
         int flow = 0;
-        int remainingFlow = MathUtils.safeInt((long) (aOptFlow * 1.25f)); // Allowed to use up to 125% of optimal flow.
-                                                                          // Variable required outside of loop for
+        int remainingFlow = MathUtils.safeInt((long) (aOptFlow * 1.25f * flowMultipliers[0])); // Allowed to use up to
+                                                                                               // 125% of optimal flow.
+        // Variable required outside of loop for
         // multi-hatch scenarios.
         this.realOptFlow = (double) aOptFlow * (double) flowMultipliers[0];
 
@@ -78,8 +79,8 @@ public class GT_MTE_LargeTurbine_SCSteam extends GregtechMetaTileEntity_LargerTu
         if (totalFlow <= 0) return 0;
         tEU = totalFlow;
         addOutput(GT_ModHandler.getSteam(totalFlow));
-        if (totalFlow != aOptFlow) {
-            float efficiency = 1.0f - Math.abs((totalFlow - aOptFlow) / (float) aOptFlow);
+        if (totalFlow != realOptFlow) {
+            float efficiency = 1.0f - Math.abs((totalFlow - (float) realOptFlow) / (float) realOptFlow);
             // if(totalFlow>aOptFlow){efficiency = 1.0f;}
             tEU *= efficiency;
             tEU = Math.max(1, MathUtils.safeInt((long) tEU * (long) aBaseEff / 10000L));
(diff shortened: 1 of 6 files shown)


koiNoCirculation:
tooltip of duranium huge turbine
3c67927f2290214a6efa5b748093026f

koiNoCirculation:
fec3f7884b5cb80c7def0464a038e4b3

Scanner result shows the optimal flow is 4915200mb/t.

koiNoCirculation:
ceca7d0628798ddb47ac99631c9b1b7e

Data of ichorium
e55a9202a57d42d15c1cd476ff6b52cb

src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/turbines/GT_MTE_LargeTurbine_SCSteam.java
```diff
@@ -55,14 +55,15 @@ public int getFuelValue(FluidStack aLiquid) {
...
-    int fluidIntoPower(ArrayList<FluidStack> aFluids, long aOptFlow, int aBaseEff, float[] flowMultipliers) {
+    long fluidIntoPower(ArrayList<FluidStack> aFluids, long aOptFlow, int aBaseEff, float[] flowMultipliers) {
         int tEU = 0;
         int totalFlow = 0; // Byproducts are based on actual flow
         int flow = 0;
-        int remainingFlow = MathUtils.safeInt((long) (aOptFlow * 1.25f)); // Allowed to use up to 125% of optimal flow.
-                                                                          // Variable required outside of loop for
+        // Variable required outside of loop for
         // multi-hatch scenarios.
         this.realOptFlow = (double) aOptFlow * (double) flowMultipliers[0];
```
chochem: I think this was just here from a copy paste error or something like that.
SC does not (!) use flowmultipliers. see https://github.com/GTNewHorizons/GoodGenerator/blob/master/src/main/java/goodgenerator/blocks/tileEntity/SupercriticalFluidTurbine.java

src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/turbines/GT_MTE_LargeTurbine_SCSteam.java
```diff
@@ -78,16 +79,16 @@ int fluidIntoPower(ArrayList<FluidStack> aFluids, long aOptFlow, int aBaseEff, f
         if (totalFlow <= 0) return 0;
         tEU = totalFlow;
         addOutput(GT_ModHandler.getSteam(totalFlow));
-        if (totalFlow != aOptFlow) {
-            float efficiency = 1.0f - Math.abs((totalFlow - aOptFlow) / (float) aOptFlow);
+        if (totalFlow != realOptFlow) {
+            float efficiency = 1.0f - Math.abs((totalFlow - (float) realOptFlow) / (float) realOptFlow);
```
chochem, line 85: which then means these changes are not needed. (just better documentation)

chochem (dismissed):
SC should not use flowmultipliers. ~not sure about SH actually.~ SH is fine

src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/turbines/GT_MTE_LargeTurbine_SCSteam.java
koiNoCirculation: Well, I don't know what original author thought, maybe we can fix that together.

koiNoCirculation: And if necessery we can add a new multiplier to SC steam for turbines.

chochem: we can talk with some more devs. But I dont think we want to buff SC steam. its already the meta for plasma.

koiNoCirculation: done.

chochem: thanks! a material specific multiplier for SC will have to wait I think until someone really wants go into balancing SC as a whole. (steelux agreed with me on discord)
I made a PR to add SC to the tooltip so it is clear to the player.

chochem (approved):

src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/turbines/GT_MTE_LargeTurbine_SCSteam.java
koiNoCirculation: Tooltip is really misleading, I wasted a whole day build fusions for duranium :(

chochem: https://github.com/GTNewHorizons/GT5-Unofficial/pull/2515

PR #851, opened 2024-03-17
add amps in hatch in one go

### LINKED ISSUE #15746: XL Turbo SC Steam turbine is lagging when used with large laser hatches
image

https://github.com/GTNewHorizons/GTplusplus/blob/71cbfcd53f7896e2fb1d304090a8a714b7510e49/src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/turbines/GregtechMetaTileEntity_LargerTurbineBase.java#L843-L845

No clue if doing `aDynamo.getBaseMetaTileEntity().increaseStoredEnergyUnits(aVoltage*ampsOnCurrentHatch, false);` instead of the full loop would break the multi in any way.

image
image

In this case, the energy hatch used is an IV 65k laser source hatch.

boubou19:
fixes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/15746

I couldn't see any obvious reason that could cause the multies to break, but i'm no expert in this part of the code.

### DIFF
--- a/src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/turbines/GregtechMetaTileEntity_LargerTurbineBase.java
+++ b/src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/multi/production/turbines/GregtechMetaTileEntity_LargerTurbineBase.java
@@ -840,10 +840,12 @@ public abstract class GregtechMetaTileEntity_LargerTurbineBase extends
             aAmpsToInject = (int) (leftToInject / aVoltage);
             aRemainder = (int) (leftToInject - (aAmpsToInject * aVoltage));
             ampsOnCurrentHatch = (int) Math.min(aDynamo.maxAmperesOut(), aAmpsToInject);
-            for (int i = 0; i < ampsOnCurrentHatch; i++) {
-                aDynamo.getBaseMetaTileEntity().increaseStoredEnergyUnits(aVoltage, false);
-            }
+
+            // add full amps
+            aDynamo.getBaseMetaTileEntity().increaseStoredEnergyUnits(aVoltage * ampsOnCurrentHatch, false);
             injected += aVoltage * ampsOnCurrentHatch;
+
+            // add reminder
             if (aRemainder > 0 && ampsOnCurrentHatch < aDynamo.maxAmperesOut()) {
                 aDynamo.getBaseMetaTileEntity().increaseStoredEnergyUnits(aRemainder, false);
                 injected += aRemainder;


Glease (changes requested):
`increaseStoredEnergyUnits(X, false)` is an "atomic" operation. You either add X amount of energy or you don't. No half-add-half-waste is ever done. Making it increase in one go will cause much greater waste when dynamo stored energy is near its capacity.

boubou19:
Yeah but i don't see the difference between using the loop or to do it in a single operation, given we don't even use the boolean returned to check if the energy has successfully been added to the hatch. So my understanding is that you'd waste the energy the same way with the loop, as the `injected` variable is unconditionally incremented no matter if the injection failed.

Glease:
There is a difference when you have 2packets of free space, but want to inject 4 packets.

boubou19:
Sure, but i don't see any logic done by the Multi to check for room before injecting to the hatch, it seems to assume the power is always considered injected by the multi no matter what, all it does it to track down how many power was added to one hatch so it can properly distribute the power to the other hatches. But in the end, if there is some power left, it's discarded, so you lose this power in the end? And the amount of amps to add to a hatch is computed the same way in both the old code and new code, so if power is discarded, it's discarded the same way, we just avoid spam calling the method to add the energy. So i don't understand your point.

Glease:
assume there is 2.5 packet worth of space, and we need to add 4 packet

before

makes 4 attempts

\# 1: added 1 packet
\# 2: added 1 packet
\# 3: fail because less than 1 packet space
\# 4: fail because less than 1 packet space

total: added 2 packet

after

make 1 attemp
\# 1: fail to add anything at all, because there is no space for 4 packet

total: nothing

boubou19:
the code used in `increaseStoredEnergyUnits`:
```java
    @Override
    public boolean increaseStoredEnergyUnits(long aEnergy, boolean aIgnoreTooMuchEnergy) {
        if (!canAccessData()) return false;
        if (getStoredEU() < getEUCapacity() || aIgnoreTooMuchEnergy) {
            setStoredEU(mMetaTileEntity.getEUVar() + aEnergy);
            return true;
        }
        return false;
    }
```

We pass it with `aIgnoreTooMuchEnergy` to `false`, so the energy will fail to be added to the buffer, if and only if, there is not room for a single EU. Which means, if there is at least room for 1 EU, it will be added to the internal storage. So worse that can happen is to have the internal buffer way above the max capacity, but we don't actually lose any energy here. In fact, we even recover some, because with the loop, once it gets full, rest is lost until the loop is done to move to another hatch.

Glease (approved):

PR #852, opened 2024-03-18
Fixes for wireless chargers

Tungsten-W-74:
Addresses Modpack issue [#14596](https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/14596)

The wireless charger stops working in single player when the player saves and quits to the title screen, then reopens the world. Closing and relaunching the client is the only way to fix this.

This happens because while the Wireless Charger object gets destroyed when the chunk is unloaded, that 'ghost' object remains in the static ChargingHelper. When the chunk is reloaded, a new wireless charger object is created, but is not added to the ChargingHelper because the 'ghost' object occupies the same position. This 'ghost' object is drained of energy but doesn't receive any energy so it runs out after a few ticks.

This fix replaces the old 'ghost' charger object with the new instance of the same charger.

### DIFF
--- a/src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/basic/GregtechMetaWirelessCharger.java
+++ b/src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/basic/GregtechMetaWirelessCharger.java
@@ -470,13 +470,13 @@ public class GregtechMetaWirelessCharger extends GregtechMetaTileEntity {
                 this.mCurrentDimension = aBaseMetaTileEntity.getWorld().provider.dimensionId;
             }
 
-            boolean mHasBeenMapped = this.equals(ChargingHelper.getEntry(getTileEntityPosition()));
-            if (!mHasBeenMapped && ChargingHelper.addEntry(getTileEntityPosition(), this)) {
-                mHasBeenMapped = true;
-            }
+            if (aTick % 20 == 0) {
+                boolean mHasBeenMapped = this.equals(ChargingHelper.getEntry(getTileEntityPosition()));
+                if (!mHasBeenMapped) {
+                    mHasBeenMapped = ChargingHelper.addEntry(getTileEntityPosition(), this);
+                }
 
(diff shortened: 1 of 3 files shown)


Tungsten-W-74:
Also just spotted another theoretical bug for MP.

The tick counter in ChargingHelper increases per playertickevent rather than per servertickevent, meaning
- only one player per second will get charged i.e. charges slower the more players are online
- while there are an even number of players online, most of them will never get charged

Will add a fix shortly.

src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/basic/GregtechMetaWirelessCharger.java
```diff
@@ -471,6 +470,7 @@ public void onPostTick(final IGregTechTileEntity aBaseMetaTileEntity, final long
                 this.mCurrentDimension = aBaseMetaTileEntity.getWorld().provider.dimensionId;
             }
 
+            boolean mHasBeenMapped = this.equals(ChargingHelper.getEntry(getTileEntityPosition()));
             if (!mHasBeenMapped && ChargingHelper.addEntry(getTileEntityPosition(), this)) {
                 mHasBeenMapped = true;
             }
```
AbdielKavash: This section is run every tick, and does some map searching for every TE. While in the grand scheme of things this is a trivial difference, I would still like to see it moved below to the "every 20 ticks" conditional. There shouldn't be any noticeable gameplay difference.

Otherwise OK, thanks for the fix!

AbdielKavash (changes requested):

src/main/java/gtPlusPlus/xmod/gregtech/common/tileentities/machines/basic/GregtechMetaWirelessCharger.java
Tungsten-W-74: Thank you for the feedback. I have made the requested change.

AbdielKavash (approved):

PR #854, opened 2024-03-19
Allow crib works for chemplant again

HoleFish:
Follow up to https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/15608
Ensure that catalyst housing will never be isolated from crib.
Remove parallel restriction from catalyst durability, as it doesn't work well with batch mode and RNG, making it meaningless.
Close https://github.com/GTNewHorizons/Dupes-Exploits-GTNH/issues/130


BlueWeabo (approved):
Looks good.

PR #697, opened 2023-07-19
Implement generic processing logic

minecraft7771:
Needs https://github.com/GTNewHorizons/GT5-Unofficial/pull/2155
Migrates most machines to generic processing logic

PR #864, opened 2024-04-07
Fix the speed of the chemplant (again)

chochem:
Yes, I fixed it a year ago. But it has since been broken again, this time in the other direction. :P

Specifically it was broken here: https://github.com/GTNewHorizons/GTplusplus/pull/697


HoleFish:
Can't we just delete this method and do a copypaste of pyrolyse oven code?

chochem:
never looked at pyro, whats better about it?

HoleFish:
Pyro one is shorter (and clear):
`setSpeedBonus(2f / (1 + coilHeat.getTier()));`
In chemplant it would be like `setSpeedBonus(2f / (1 + mCoilTier));` or smth

chochem:
done.

HoleFish (approved):

PR #867, opened 2024-04-15
Temporarily add Multi-Amp support to some GT++ Multis via upgrade chip

MeiTianyou:
This adds an upgrade chip that allows the Extruder, LPF, Press, Forge Hammer, Mixer, and Cutting Machine to use multi-amp hatches (but not lasers). It also doesn't allow overclocking past the energy hatch tier. I have seen several late game players say that this would make their lives better while more interesting, funner, gameplay is developed.

I really think it would be beneficial to poll lategame players to see if this is something that is desired before a conversation about balancing the chip recipe. It's pointless for everyone to argue about just exactly how terrible the recipe should be, if it turns out only like three people want this feature. In any case, I did my best to implement it in an easily reversible fashion.

This is a draft, as there are some things I would like help with.  

- Maybe the most obvious thing is that the chip really should have a multi-line tooltip, but I have no idea how to make that happen within the gtpp framework
- There should be a quest prepared for if this gets added
- General review, because frankly I wanted to get a working thing asap
- It needs more testing than I can do alone

Staffi reminded me I made a poorly edited paint collage 
ma


NotAPenguin0:
If anything this should at least be gated to UXV in my opinion, but aside from that I'm not sure this is something that will ever get in.

C0bra5:
> If anything this should at least be gated to UXV in my opinion, but aside from that I'm not sure this is something that will ever get in.

As it is, you don't get recipe tier skips, just more power budget for parallels, so IMO, it's fine because the machines are literal dirt, and all this does is reduce machine spam, which is something that starts to kick in around UIV. Not to mention, it's also less power efficient than machine spam since the OCs will increase the overall power use (unless you can't reach the cap like with energized tesseracts), but even then, it's still just reducing the number of machines you have to slap down since you'd have to slap down more machines anyway to achieve the same productivity ratio.

Steelux8:
I am opposed to this, as I've said before, because these multis were not designed to be used in this tier, and this change incentivizes that usage. I am not an endgame player, but I do understand how easy it effectively is to add this short-term solution to the significant problem that exists. That said, if such a rushed solution is acceptable, I'd rather vouch for new simple multis, perhaps similar to the GT++ ones, to be added specifically to solve this problem, instead of reusing IV multis. If there is no need to gauge balance, design and such things, then it would be very easy and quick to add dedicated multis for this purpose.

evgengoldwar:
> I am opposed to this, as I've said before, because these multis were not designed to be used in this tier, and this change incentivizes that usage. I am not an endgame player, but I do understand how easy it effectively is to add this short-term solution to the significant problem that exists. That said, if such a rushed solution is acceptable, I'd rather vouch for new simple multis, perhaps similar to the GT++ ones, to be added specifically to solve this problem, instead of reusing IV multis. If there is no need to gauge balance, design and such things, then it would be very easy and quick to add dedicated multis for this purpose.

You just need to rebalance and add Twist-Space-Technology-Mod

Steelux8:
Rebalancing the Twist mod alone is a large amount of work, and this would take some time even if it started being done right now.

evgengoldwar:
> Rebalancing the Twist mod alone is a large amount of work, and this would take some time even if it started being done right now.

Twist solves all high tier problems, just rebalance it a little and that's it

NotAPenguin0:
It's really not that simple, as Steelux said it would take a lot of time and effort to get it to a good state.

C0bra5:
This PR is better than doing nothing, which has been the status quo for a while now...

evgengoldwar:
> It's really not that simple, as Steelux said it would take a lot of time and effort to get it to a good state.

Yes, it costs some effort, but still it’s nothing compared to doing nothing or writing all these multis from scratch

Steelux8:
Doing something that is poor in quality is not always better than doing nothing. Sometimes, a new introduction causes problems down the line that makes it so it was better to have done nothing in the first place.

NotAPenguin0:
Should also note that "temporary solutions" have a tendency to become permanent. If the development of a replacement system drags out for long then people are not going to want to remove this feature because of backwards compatibility issues, etc.

boubou19:
It's not in line with what we want for the pack for now, so i'm afraid this will never make it into the pack.

C0bra5:
> I am opposed to this, as I've said before, because these multis were not designed to be used in this tier, and this change incentivizes that usage. 

I'm sorry, but this argument does not hold to scrutiny. There is no available alternative to these machines at this moment; the end-game content has been in the game for literal years. If the team didn't want players to be using machines unlocked in IV with end-game content, it stands to reason that an alternative would have been created and implemented a long time ago.

> That said, if such a rushed solution is acceptable, I'd rather vouch for new simple multis, perhaps similar to the GT++ ones, to be added specifically to solve this problem, instead of reusing IV multis. If there is no need to gauge balance, design and such things, then it would be very easy and quick to add dedicated multis for this purpose.

The fact that the upgrade chip requires materials that can only be obtained when a player reaches UIV pretty much already addresses this point.

A good middle ground I think we could propose would be to have a separate chip for each machine, Each chip would require the UIV (or tier ends up being decided on) equivalent of the controller recipe or maybe something similar with an active transformer and and tesseract needed to make a wireless energy hatch. That way, we can easily attach it to the existing system, and the cost would remain somewhat relatively balanced for the tier that the chip is made in. 

If anything this PR offers a tradeoff to the players:
- Increase overall power usage since you can now perform recipe-time-only OCs on recipes that use a full amp of power and have a high parallel count (which is most recipes at the end game).
- Spam more machines but save on overall energy costs, since you aren't performing extraneous, overclocks.

This trade-off is similar to the existing parallel machine vs overclock dilemma.

Given the fact that MuTEs will replace all gt++ multis and eventually sunset them, there is no need to worry about the future impact players will have to migrate off of them.

> It's not in line with what we want for the pack for now, so I'm afraid this will never make it into the pack.

So I'm to assume that the team wants players to just spam more machines, then.

evgengoldwar:
> C0bra5

The good middle ground is Twist, which completely changes the high tier for the better.

Dream-Master:
Why not add new multis instead of recycling old multis and putting them on steroids? Sure it's easier to find a temporary solution now, but at the end of the day we live with a lot of temporary and duct tape solutions.

Ya9yu:
I agree with the messages above, we don’t need temporary solutions because there are already too many of them, but we already have a Twist addon, which just needs to be well balanced, this will be a very good solution, as it can remove this problem, and also add variability high tires

MeiTianyou:
> Why not add new multis instead of recycling old multis and putting them on steroids? Sure it's easier to find a temporary solution now, but at the end of the day we live with a lot of temporary and duct tape solutions.

Simply because people are not allowed to do that

noobyaran9:
PAs used to give 256 or 1024 parallel when using UIV or UMV machines in them.
This looks to be a better solution than allowing multi-A hatches on GT++ machines.
Not sure why this was removed at one point.

NotAPenguin0:
This is still in the game using the downtiering option.

noobyaran9:
> This is still in the game using the downtiering option.

Only upto 64 parallels, anything higher is removed.

NotAPenguin0:
That's simply not true, using 64 UMV machines with downtiering on will run 256 UIV machines instead. When was this removed according to you?

noobyaran9:
> That's simply not true, using 64 UMV machines with downtiering on will run 256 UIV machines instead. When was this removed according to you?

Sorry, my mistake.
What I meant was:
64 UEV -> 256 UHV
64 UIV -> 1024 UHV
64 UMV -> 4096 UHV

Now its only able to do 1 down tier instead of able to down tiering to UHV tier.

Dream-Master:
> > Why not add new multis instead of recycling old multis and putting them on steroids? Sure it's easier to find a temporary solution now, but at the end of the day we live with a lot of temporary and duct tape solutions.
> 
> Simply because people are not allowed to do that

Why so? Since Mutes will be delayed you are free to add new mtes. You not come up with a concept .

MeiTianyou:
> > > Why not add new multis instead of recycling old multis and putting them on steroids? Sure it's easier to find a temporary solution now, but at the end of the day we live with a lot of temporary and duct tape solutions.
> > 
> > 
> > Simply because people are not allowed to do that
> 
> Why so? Since Mutes will be delayed you are free to add new mtes. You not come up with a concept .

First I heard of that the ban on mtes is gone. Bless

C0bra5:
> First I heard of that the ban on mtes is gone. Bless

it's official now
image

Caedis:
Still does not mean any multi will be added. They have to be approved before dev work starts.

Dream-Master:
I mean that we just lift the ban of mtes but a written concept have to be shown first before codeing starts.

PR #872, opened 2024-04-19
Fix Thermal Boiler accepting energy hatches.

AbdielKavash:
Makes the Thermal Boiler not form with an energy hatch. It is a boiler, it does not need energy. This subsequently prevents it from trying to overclock and generate incorrect amounts of steam. Mentioned here: https://discord.com/channels/181078474394566657/603348502637969419/1230902328203280435, thanks to @Reflex18 for the report.


chochem (approved):
lgtm

PR #875, opened 2024-04-22
Fix steam multi bonuses and document them

chochem:
They were broken by the GPL implementation here https://github.com/GTNewHorizons/GTplusplus/pull/697.

This PR restores the old values and makes them clear in the tooltip.

example calculations:
1 crushed ore in a singleblock (non-HP) steam macerator takes 40 seconds and uses 4L/t steam = 3200 steam total.
The multi macerates 8 crushed ore in 30 seconds for 22L/t steam = 13200 steam total or 1650 steam per ore.
So you get roughly 2x efficiency and more than 10x speed compared to the singleblock.


Ethryan (approved):
LGTM

PlayfulPiano:
and so ends the hilarity of 4x efficient squashers

PR #878, opened 2024-05-07
Fixed the controller texture for the turbines.PlayerCache is not called anywhere?

Pilad:
Fixed the controller texture for the turbines. PlayerCache is not called anywhere?
изображение


Pilad:
It is also necessary to correct the texture of the block in NEI. But it doesn’t seem to interfere.
изображение

src/main/java/gtPlusPlus/core/util/player/PlayerCache.java
NotAPenguin0, line 1: In the future try to put separate changes in a separate PR

NotAPenguin0 (approved):

PR #879, opened 2024-05-10
Add RawOre items and processing 

Dream-Master:
depends on: https://github.com/GTNewHorizons/GT5-Unofficial/pull/2502


Ethryan (approved):
Approved, but needs a dep update before merging.

PR #880, opened 2024-05-10
Add textures for numbered bio and breakthrough circuits

GDCloudstrike:
This PR adds textures for all the number configurations of bio and breakthrough circuits, similar to the regular programmed circuits.

image


NotAPenguin0 (approved):
Based

PR #881, opened 2024-05-14
Cleaning!

Connor-Colenso:
- Remove a ton of unused textures (some double checking on this would be helpful, nothing obvious in NEI).
- Removed debug scanner, doesn't seem to do anything useful.
- Changed name of mobessence to Mob Essence.
- Removed xpjuice fluid (this is some useless GT++ variant as far as I can tell).
- Clean up a bunch of configs.
- Clean up tooltips and remove some old ones.
- Remove sick blazes, I couldn't even get these to spawn?

Other misc stuff as well but those are the main points I think


Dream-Master:
@Connor-Colenso did you add back fluorine ? Was removed with the last cleanup . Gt++ had a own xp recipe not sure this is vanished when you remove the xp fluid ?

chochem (dismissed):
1. already discussed on discord, but just to put it here too: the blazes work perfectly fine. both the egg and natural spawns. please revert that part.
2. see comment about xp conversion recipes

for the textures I am not sure how best to check either, maybe there is something in the start-up log if one of the textures is now missing for something.

chochem:
> @Connor-Colenso did you add back fluorine ? Was removed with the last cleanup . Gt++ had a own xp recipe not sure this is vanished when you remove the xp fluid ?

the xpjuice is added by openblocks and still will be after this. this was just a fallback for when openblocks is missing.

src/main/java/gtPlusPlus/core/recipe/RECIPES_GREGTECH.java
```diff
@@ -565,12 +565,6 @@ private static void laserEngraverRecipes() {
     }
 
     private static void breweryRecipes() {
-        GT_Values.RA.stdBuilder().itemInputs(GT_Utility.getIntegratedCircuit(14))
```
chochem, line 568: I dont see any reason to remove these conversions.
can do it directly without 'enchanting utils' of course

chochem: of course even without enchanting utils that will still need some check either that the fluid exists or that openblocks is loaded, to avoid a crash in gt++ devenv

Connor-Colenso:
The sick blazes are back, although they do nothing. @chochem

chochem (approved):
nobody else found any issues, so lets go with it.

